Issue #11, opened 2026-07-02
Phase 2: Isolated correctness fixes (3 math bugs)


neuromechanist:
Part of epic #9. Depends on Phase 1 gate.

## Scope (each with a regression test vs Fortran)
1. Per-source mixture factorization: compute `sum_i log sum_k alpha_{k,i} f_k(y_{i,t})`, not sum-over-sources-then-logsumexp. Fix `_compute_gg_log_pdf_vectorized` / `compute_log_likelihood` in `amica_torch.py` (and v2 fallback).
2. Per-source alpha: use `alpha[k, :]` per source, not `alpha_k.mean()` (`amica_torch.py:187`, `amica_torch_v2.py:259`).
3. LL normalization: divide by `n_samples * nw` and drop the extra `n_samples*log_det_W` term to match Fortran units (`amica17.f90:1866`).

## Done when
All three fixes land with tests locking them to Fortran values; Phase 1 baseline report shows improved per-channel LL scale (~ -3.4 range) and no regression in correlation.

neuromechanist:
Lead from Phase 1 review (PR #14): the legacy-NumPy Newton NaN (behind test_amica.py::test_full_amica and test_pyAMICA.py::test_full_pipeline) traces to a divide-by-zero at `pyAMICA/pyAMICA.py:747` and `:751` (`dmu = updates['dmu'] / updates['dalpha']`, `dbeta = ... / updates['dalpha']`) when a mixture component's `dalpha` collapses to zero. Looks like a clean guard/clamp fix in Phase 2 scope (Fortran floors these with epsilon), not something needing the Phase 3 rewrite. Add a regression test that exercises the collapse case.

neuromechanist:
Phase 2 (isolated correctness fixes) delivered via epic #9, on main in PR #28. Closing.

Issue #9, opened 2026-07-02
Epic: PyTorch backend parity with Fortran AMICA (natural-gradient EM rewrite)


neuromechanist:
## Goal

Bring the PyTorch backend to numerical parity with the Fortran reference (`amica15mac`), and fix the memory/scaling problems that come from the current Adam+autograd framing.

## Root cause

The default torch backend reframes AMICA as "minimize NLL with Adam over reparameterized tensors," but AMICA is a fixed-point EM / natural-gradient method with closed-form sufficient-statistic updates. Adam on the reparameterized surface follows a different trajectory and converges to a different fixed point. The NumPy `pyAMICA.py` is a faithful port and is the spec.

See design review: `.context/decisions/0001-torch-backend-natural-gradient-em.md` and the "2026-07-02 design review" section of `.context/research.md`.

## Concrete defects
- Swapped mixture factorization (`amica_torch.py:176-233`): sources summed before per-source mixture reduction.
- Per-source `alpha` collapsed to a scalar mean (`amica_torch.py:187`).
- LL normalized by `n_samples` instead of `n_samples * nw` (`amica_torch.py:205`).
- Newton double-steps and NaNs at `newt_start`.
- `pinv(A.T).T` per forward; float32 default; Python loops in the hot path; no sample blocking.

## Phases
1. Validation harness & test repair (foundation gate)
2. Isolated correctness fixes (3 math bugs)
3. Natural-gradient EM rewrite (new module, swap at parity)
4. Newton + adaptive PDF + outlier rejection

## Definition of done
Component correlation vs Fortran > 0.95 on the sample EEG data; per-channel LL matching Fortran within tolerance; validation harness green as the parity gate.

neuromechanist:
Phase 3 merged (PR #20). Natural-gradient EM backend (`AMICATorchNG`) landed opt-in via `backend="ng"`; default unchanged. Sufficient statistics proven to match the NumPy reference to 1e-14; end-to-end correlation 0.69 (up from 0.43). The >0.95 target is carried into Phase 4 (Newton) as a strict xfail. Phases 1-3 complete; Phase 4 (#13) is next.

Issue #2, opened 2025-01-20
Handle and verify tests


neuromechanist:
I am working on some decent tests. The main objective should be to maximize the output results of the original package. Still, some internal tests are necessary to ensure that the core implementation as well as different features of the package, such as rejection, Newton methods, different PDF types, multiple models, etc work as expected. I see this verification to be one of the longest works I need to do

neuromechanist:
Done / superseded. Core features now have real-data tests: parity + Newton posdef (#24), outlier rejection, all five PDF families (#26, `test_ng_pdf_families.py`), multi-model (#27). The end-to-end real-data suite was stood up in Phase 1 (#10, 35 passed / 6 documented xfail). Remaining specific coverage gaps are tracked in #15 (save/load, plot_components) and #61 (edge cases + stability regressions).

PR #19, opened 2026-07-02
Phase 2: Isolated correctness fixes (mixture factorization, alpha, LL scale)

neuromechanist:
## Summary
Fixes the three isolated math bugs in the current Adam-based torch backend, plus the legacy-NumPy Newton divide-by-zero. Surgical fixes only; the architecture rewrite is Phase 3.

Closes #11
Part of epic #9

## Fixes
- **Mixture factorization** (`amica_torch.py`, `amica_torch_v2.py` fallback): per-source `logsumexp` over mixtures THEN sum over sources, instead of summing sources before the mixture reduction.
- **Per-source alpha**: use `alpha[k, :]` per source, not `alpha_k.mean()`.
- **LL normalization**: divide by `n_samples * n_sources` (Fortran `amica17.f90:1866`). Jacobian handled via `for_training` flag — excluded from the reported/Fortran-parity LL (Fortran omits `log|det W|` too), included in the training loss gradient (removing it entirely destabilized training: LL went positive, A error 15.3).
- **Legacy Newton divide-by-zero** (`pyAMICA.py`): floor `dalpha` at 1e-10 before dividing; eliminates the NaN.

## Result (real sample data, seed 42)
| | 100 iters before | 100 iters after |
|---|---|---|
| Fortran LL | -3.4108 | -3.4113 |
| PyTorch LL | -43.80 | **-0.967** |
| LL ratio | 12.8x | **0.28x** |
| Mean corr | 0.4313 | 0.4312 |

LL scale now comparable to Fortran (Known Issue #2 resolved). Correlation unchanged as expected — the Adam-vs-natural-gradient trajectory divergence is Phase 3's scope.

## Test plan
- New `test_correctness_fixes.py`: 5 tests, real data, hand-computed numpy/scipy reference for the GG-mixture LL; asserts per-source factorization, per-source alpha effect, and LL scale.
- Full suite: 40 passed, 6 xfailed (verified by team lead).
- xfail reasons updated where the NaN fix changed the failure mode to the correlation gap.

Note: first commit includes ruff-format churn on two never-formatted files (pre-commit hook).

### DIFF
--- a/pyAMICA/pyAMICA.py
+++ b/pyAMICA/pyAMICA.py
@@ -757,18 +757,27 @@ class AMICA:
         # Update mixture weights
         self.alpha = updates["dalpha"] / np.sum(updates["dalpha"], axis=0)
 
+        # dalpha is the per-component responsibility mass and is also used
+        # below as the divisor for dmu/dbeta/drho. When a mixture component's
+        # responsibility collapses to (near) zero -- e.g. an outlier
+        # component with no assigned samples -- dividing by it directly
+        # produces NaN and poisons mu/beta/rho for that component. Floor it
+        # with a small epsilon, matching the epsilon-floor pattern used
+        # elsewhere in this codebase (e.g. invsigmin/invsigmax clipping).
+        dalpha_safe = np.maximum(updates["dalpha"], 1e-10)
+
         # Update component means
-        dmu = updates["dmu"] / updates["dalpha"]
+        dmu = updates["dmu"] / dalpha_safe
         self.mu += self.lrate * dmu
 
         # Update scale parameters
-        dbeta = updates["dbeta"] / updates["dalpha"]
+        dbeta = updates["dbeta"] / dalpha_safe
         self.beta *= np.sqrt(1 + self.lrate * dbeta)
         self.beta = np.clip(self.beta, self.invsigmin, self.invsigmax)
 
         # Update shape parameters
         if not np.all(self.rho == 1.0) and not np.all(self.rho == 2.0):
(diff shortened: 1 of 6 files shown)


neuromechanist:
## Review outcome (Sonnet code-reviewer + team-lead verification)

**Confirmed correct — keep as-is:**
- Bug 1 (per-source mixture factorization): traced line-by-line against `amica17.f90:1313-1360`; `logsumexp(dim=0).sum()` matches Fortran's per-source reduction then sum-over-sources.
- Bug 2 (per-source alpha): `alpha_k.unsqueeze(1)` correct; no downstream `.mean()` collapse remains.
- Legacy `dalpha` epsilon floor: sound narrow guard, eliminates the Newton NaN.
- The new `test_correctness_fixes.py` numpy reference is a genuine independent re-derivation for bugs 1 and 2.

**Reworked before merge (bug 3):** The `for_training` Jacobian split was based on a wrong premise (mine) that Fortran omits `log|det W|` from the reported LL. Verified against source: Fortran INCLUDES it — `amica17.f90:975-980` computes `log|det W|` via QR of W, and `:1273` seeds `Ptmp = Dsum + log(gm) + sldet` (unmixing + sphering log-determinants) before the density loop, flowing into the printed `LL` at `:1866`. Fix: include `log|det W|` unconditionally in the reported LL and add the sphering log-det `sldet`; drop the `for_training` split (which also silently gave the Newton/optimizer call sites a Jacobian-less objective). Test bound tightened toward Fortran's ~ -3.4.

Canonical docs corrected on the epic branch (`research.md`). Bugs 1/2 stand; bug 3 rework in progress.

Issue #13, opened 2026-07-02
Phase 4: Newton + adaptive PDF + outlier rejection


neuromechanist:
Part of epic #9. Depends on Phase 3.

## Scope (fan out independent sub-tasks)
- Newton: port H matrix from sigma2/kappa/lambda and baralpha (`pyAMICA.py:666-694`); Fortran-matched lrate/newton ramp; delete the autograd Newton that double-steps.
- Adaptive PDF selection ported into the new backend (density-parameter driven, not the ad-hoc kurtosis heuristic).
- Outlier rejection (`do_reject`, reject_data) ported from NumPy.

## Done when
Newton path is stable (no NaN at newt_start) and improves convergence vs natural-gradient-only; adaptive PDF and rejection match NumPy behavior; parity gate still green.

neuromechanist:
Phase 4 delivered via epic #9: Newton (positive-definite, 0 fallbacks) and outlier rejection (do_reject) are implemented on AMICATorchNG; adaptive-PDF selection split out to #26. On main in PR #28. Closing.

Issue #21, opened 2026-07-03
NG backend converges to a different fixed point than Fortran (~0.68 corr, blocks >0.95 gate)


neuromechanist:
## Summary

The `AMICATorchNG` backend plateaus at ~0.68 Hungarian-matched component correlation vs the
Fortran binary (LL ~ -3.47 vs Fortran -3.41) on the sample EEG data, and this is **independent of
Newton**. Phase 4 (#13) ported Newton, adaptive-PDF, and outlier rejection into the NG backend and
verified them to float64 precision against the (Fortran-corrected) NumPy reference, but they only
accelerate convergence to the NG backend's own fixed point. This blocks the epic (#9) definition of
done (>0.95 correlation), so `test_end_to_end_correlation_vs_fortran` remains xfail.

## Evidence (all 100 iters vs `amica15mac`, seed 42; correlation ~0.65-0.69 in every case)

- data-space natural gradient vs Fortran source-space `g^T b` (b=Wx): both ~0.68.
- gradient-style `mu`/`beta` updates vs Fortran's exact EM fixed-point updates
  (`mu += dmu_numer/dmu_denom`, no lrate, rho-dependent denominators; `sbeta *= sqrt(...)`): both ~0.68.
- Newton on vs off: both ~0.68 (Newton is positive-definite and active every iteration when on).
- Whitening: NG PCA-whitening vs a symmetric-ZCA rotation made it worse (0.44), so it is a real
  fixed-point plateau, not a basis mismatch.

Because no update-rule change moves the plateau, the remaining suspects are outside the update
equations:

1. **Initialization mismatch.** NG seeds `A`/`mu`/`beta` via numpy `RandomState` (matched to the
   NumPy port), but the Fortran binary uses its own RNG/init, so the two start from different
   points of a non-convex objective.
2. **Iteration count.** Fortran's default `max_iter` is 2000; the gate test uses 100.
3. A subtler E-step / responsibility difference.

## Suggested next steps

- Replicate Fortran's exact initialization (seed -> starting `A`/`mu`/`beta`/`rho`) so both start
  from the same point, then re-measure.
- Compare NG vs Fortran per-iteration (`W`, LL, and a couple of components) to localize where the
  trajectories diverge.
- Consider adopting Fortran's exact EM `mu`/`beta`/`alpha` updates and source-space natural
  gradient in the NG backend regardless (they are more faithful and more numerically stable at
  lrate=0.05, even though they do not close the gap alone).

Part of epic #9.

[...]

PR #22, opened 2026-07-03
Phase 4: Newton + adaptive PDF + outlier rejection

neuromechanist:
## Summary

Ports the Fortran Newton preconditioner, adaptive-PDF (rho) confirmation, and outlier rejection
into the natural-gradient backend `AMICATorchNG`, and adds Fortran-style learning-rate control.
All three features are implemented and unit-parity-verified against the (Fortran-corrected) NumPy
reference to float64 precision.

**Important:** this does NOT yet close the epic's >0.95 correlation gate. The NG backend converges
to a different fixed point than Fortran (~0.68 correlation, LL -3.47 vs -3.41), and Newton only
accelerates convergence to that same NG fixed point. The root cause is upstream of Newton and is
tracked separately in #21. The end-to-end gate test is therefore kept as xfail with an accurate
reason.

Closes #13. Part of epic #9.

## What changed

- `torch_impl/amica_torch_ng.py`
  - Newton preconditioner: score-based curvature stats (sigma2/lambda/kappa), vectorized
    per-source-pair 2x2 direction solve, positive-definiteness fallback to natural gradient,
    newtrate-capped ramp. Matches `amica17.f90`.
  - Outlier rejection: per-sample log-likelihood threshold (`mean - rejsig*std`), one-directional
    good-sample mask, good-count normalization of `gm` and reported LL.
  - Learning-rate control: replaces the ad-hoc backtracking with Fortran's convergence logic
    (anneal on LL decrease; ratchet the lrate ceiling and newtrate after `maxdecs` decreases). This
    fixes a pre-existing divergence where the natural gradient drifted the LL down at lrate=0.05.
  - Extracted a shared `_forward` E-step helper (used by block updates and the rejection LL).
- `pyAMICA.py`: corrected the NumPy reference Newton math to Fortran spec (score `fp` not the
  density derivative; `dsigma2` accumulated once per source; `sbeta^2` on kappa; baralpha-weighted
  `mu^2` term in lambda; positive-definiteness fallback), so it is a valid Newton parity target.
- `amica.py`: `do_newton` now plumbs through `backend="ng"` instead of raising.

## Test plan

- `test_newton_stats_match_numpy_reference` - finalized sigma2/lambda/kappa vs NumPy, <1e-8.
- `test_newton_mstep_matches_numpy_reference` - one full Newton M-step A-update vs NumPy, <1e-8.
- `test_newton_direction_matches_formula` - direction formula + posdef guard on controlled input.
- `test_rejection_shrinks_good_sample_set` - real-data rejection drops samples, stays finite.
- `test_end_to_end_correlation_vs_fortran` - xfail (fixed-point gap, #21).
- Full fast suite green: 46 passed, 4 xfailed, 0 failures. Real sample EEG + Fortran binary only.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ enhanced_test_output/
 .coverage
 coverage.xml
 htmlcov/
+
+# Temp output from NG e2e test (Fortran run artifacts)
+pyAMICA/tests/torch_tests/_ng_e2e_tmp/
--- a/pyAMICA/amica.py
+++ b/pyAMICA/amica.py
@@ -119,10 +119,10 @@ class AMICA:
         do_sphere : bool, default=True
             Whether to sphere (whiten) the data
         do_newton : bool, default=False
-            Whether to use Newton optimization (experimental). Only supported
-            by ``backend="torch"``; passing ``do_newton=True`` with
-            ``backend="ng"`` raises, since Newton is not yet ported to the
-            natural-gradient backend (Phase 4).
+            Whether to use Newton optimization. Supported by both backends.
+            For ``backend="ng"`` this enables the Fortran-parity Newton
+            preconditioner; tune it via ``newt_start``/``newtrate`` in
+            ``**kwargs``.
         output_dir : str, optional
             Directory for debug output (only used if debug=True). Only
             supported by ``backend="torch"``.
@@ -158,11 +158,6 @@ class AMICA:
             device = self.device
 
         if self.backend == "ng":
-            if do_newton:
-                raise ValueError(
-                    "do_newton=True is not supported by backend='ng' yet "
-                    "(Newton is Phase 4, not ported to AMICATorchNG)."
-                )
             if debug:
                 raise ValueError(
                     "debug=True (Fortran-style output) is not supported by "
@@ -181,6 +176,7 @@ class AMICA:
                 lrate=lrate,
                 do_mean=do_mean,
                 do_sphere=do_sphere,
+                do_newton=do_newton,
                 device=device,
                 **kwargs,
             )
--- a/pyAMICA/amica.py
+++ b/pyAMICA/amica.py
@@ -119,10 +119,11 @@ class AMICA:
         do_sphere : bool, default=True
             Whether to sphere (whiten) the data
         do_newton : bool, default=False
-            Whether to use Newton optimization. Supported by both backends.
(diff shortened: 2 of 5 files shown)


neuromechanist:
## /review-pr summary (3 agents, Sonnet)

Ran code-review, test-coverage, and silent-failure reviewers on the diff. No findings on the (intentional, tracked) >0.95 gate. Findings addressed in ec84388:

**Correctness / parity**
- Rejection now uses the pre-update per-sample log-likelihood (Fortran computes `loglik` in the E-step before `update_params`); previously it recomputed from post-update parameters, a spec deviation.
- Ported the NumPy reference's `dalpha` epsilon floor to the NG backend (`do_reject` makes a zero-responsibility divide more likely).

**Robustness (no silent failures)**
- Validate `rejint`/`rejsig`/`maxrej` at construction; raise a clear `ValueError` when a degenerate/NaN LL would empty the good set, instead of the opaque downstream `None`/`TypeError` crash the reviewers reproduced.
- Newton fallback is no longer silent: `n_newton_fallbacks` counter + `logger.warning` (mirrors Fortran's print), `stop_reason` attribute, and the lrate-floor stop promoted to `warning`.

**Docs**
- Fixed stale/overclaiming docstrings: the lrate-control params (this PR implements the convergence decay the old text said was absent), `do_newton` wiring (only functional for `backend="ng"`), and two test docstrings that claimed to exercise the positive-definite Newton path (on this data the curvature is not positive definite, so it always falls back — confirmed by the reviewers).

**Coverage added**
- Positive-definite Newton M-step composition (forced posdef, independent recomputation of `A + lrate*(A@H)` + rescale).
- `rho`=1.0/2.0 boundary stat parity (the `_score` special cases).
- Multi-model finiteness/shape (NumPy parity is intentionally invalid for `n_models>1` due to the Jacobian-in-`v` deviation).
- Degenerate-LL rejection error, reject param validation, and LL good-count normalization.

Full fast suite: 52 passed, 4 xfailed, 0 failures. Reviewers explicitly noted the diagonal Newton `lambda` division is unguarded but matches Fortran (`Wtmp(i,i)=dA(i,i)/lambda(i)`, no check), so kept Fortran-faithful; the NaN-LL break + enriched logging make it diagnosable.

The unmet >0.95 gate is a real, deeper NG-vs-Fortran fixed-point discrepancy (issue #21), not a Newton bug; the gate test stays xfail.

Issue #29, opened 2026-07-03
Address PR #28 review: NG backend polish + doc coherence


neuromechanist:
Follow-ups from the PR #28 (epic close-out) holistic review. No blockers to the epic, but the "no technical debt carried forward" policy wants these cleaned up before/with the epic merge.

## Important
- [ ] `AMICA(backend="ng")` crashes on default config on Apple Silicon: `fit()` auto-selects MPS, but `AMICATorchNG` defaults to float64 (Fortran parity), which MPS rejects. Fall back to CPU when MPS is auto-selected for a float64 NG run; document.
- [ ] No test for the `amica.py` `backend="ng"` wrapper (constructor guards, debug/output_dir guards, save/load NotImplementedError, real-data fit). Add wrapper test.

## Docs coherence
- [ ] AGENTS.md env-setup stale re: UV migration; Architecture Map / Key Files omit `amica_torch_ng.py` / `AMICATorchNG`.
- [ ] ADR `.context/decisions/0001-torch-backend-natural-gradient-em.md` still "Status: proposed" (+ README index); should be Accepted.
- [ ] Two xfail tests reference "epic #9" as the blocker; repoint to concrete follow-up issues.
- [ ] `PROGRESS_SUMMARY.md` / `FEATURE_PARITY.md` carry stale pre-epic numbers.

## Suggestions
- [ ] `amica.py` class docstring `model_ : AMICATorch` should note `AMICATorchNG` for `backend="ng"`.
- [ ] `pyAMICA/__init__.py` does not re-export `AMICATorchNG`.

## Related
- Removing the parked `AMICATorchV2` once NG is the de-facto main backend (separate scope).

neuromechanist:
Done and on main: fixes landed via PR #33 (into the epic) and reached main with the epic close-out PR #28. Follow-ups split out as #30, #31, #32, and epic #34.

Issue #32, opened 2026-07-03
Remove parked AMICATorchV2 and promote backend='ng' to default


neuromechanist:
Now that `AMICATorchNG` (`backend="ng"`) reaches Fortran parity (issue #24), the earlier
`AMICATorchV2` (`torch_impl/amica_torch_v2.py`) is parked/superseded and only adds maintenance
weight. Once `backend="ng"` becomes the de-facto main backend, clean up the superseded paths.

## Scope
- [ ] Remove `pyAMICA/torch_impl/amica_torch_v2.py` (`AMICATorchV2`) and its exports/references.
- [ ] Remove any now-dead tests/docs pointing at V2.
- [ ] Decide the default backend: promote `backend="ng"` to the `AMICA` default (currently
      `backend="torch"`), or make `"ng"` the only backend and retire the basic `AMICATorch`
      (depends on resolving the legacy mixture M-step bug #31).
- [ ] Update ADR 0001 / AGENTS.md / FEATURE_PARITY.md / PROGRESS_SUMMARY.md accordingly.

## Depends on / relates
- Adaptive PDF on NG: #26
- Multi-model partition matching: #27
- Legacy basic-backend bugs: #30 (CLI save/load), #31 (mixture M-step)

Deliberately out of scope for the epic close-out (PR #28); this is a follow-up so the removal gets
its own focused change.

Issue #31, opened 2026-07-03
GaussianMixtureICA.fit_em log-likelihood decreases across iterations


neuromechanist:
`GaussianMixtureICA.fit_em` (`torch_impl/mixture_models.py`, used by the basic `backend="torch"`
path) has an M-step bug: the log-likelihood decreases across EM iterations instead of ascending.

Surfaced by the xfail `pyAMICA/tests/torch_tests/test_torch_components.py::test_em_fitting` (strict
xfail). Previously mis-attributed to "epic #9"; the epic delivered the natural-gradient backend
(`AMICATorchNG`), which does not touch this legacy mixture module. Fix the M-step update (or retire
this module if `backend="torch"` is removed in favor of `backend="ng"`), then un-xfail the test.

neuromechanist:
Resolved by #32 (PR #35): the buggy `GaussianMixtureICA.fit_em` M-step lived in `torch_impl/mixture_models.py`, which was deleted when the basic `AMICATorch` backend was removed. `AMICATorchNG` (now the sole backend) uses the Fortran-parity exact-EM mixture updates and does not have this bug. The strict-xfail test that surfaced it was removed with the backend.

Issue #30, opened 2026-07-03
Legacy CLI save/load format mismatch (.npy vs Fortran binary)


neuromechanist:
`AMICA._optimize()` / `save_results` in `pyAMICA.py` write results as `.npy` files, but
`loadmodout()` expects raw Fortran-format binary files with no extension. The round-trip raises
`FileNotFoundError` for 'W' before any correlation check runs.

Surfaced by the xfail `pyAMICA/tests/test_sample_data.py::test_sample_data_cli` (strict xfail).
Previously mis-attributed to "epic #9"; it is an independent legacy-CLI bug. Fix the save/load
format contract (or converge on one format), then un-xfail the test.

PR #33, opened 2026-07-03
Address PR #28 review: NG backend polish + doc coherence

neuromechanist:
Closes #29. Applies the PR #28 (epic close-out) review findings into the epic so #28 carries them.

## Important
- Fix `AMICA(backend="ng")` crashing on its own default config on Apple Silicon: `fit()`
  auto-selected MPS, but `AMICATorchNG` defaults to float64 (Fortran parity), which MPS rejects.
  When the device is not pinned and resolves to MPS for a float64 NG run, fall back to CPU (CUDA is
  unaffected). Users can still opt into MPS with `dtype=torch.float32, device="mps"`.
- Add a real-data wrapper test for `backend="ng"` (`test_amica_ng_wrapper.py`): constructor guard,
  debug/output_dir guards, save/load `NotImplementedError`, the MPS->CPU device fallback, and a
  small real-data fit + transform/get_mixing/get_unmixing. No synthetic data.

## Doc coherence
- AGENTS.md: backend map now lists `amica_torch_ng.py` / `AMICATorchNG`; UV migration marked
  complete (matches `pyproject.toml`/`uv.lock`/plan.md); follow-ups linked.
- ADR `0001` accepted (+ README index) with an outcome note (issue #24 parity result).
- `FEATURE_PARITY.md` / `PROGRESS_SUMMARY.md`: replaced stale pre-epic metrics (~0.78 corr, ~13x
  LL) with the NG parity numbers (LL ~ -3.40, corr ~0.997); the "PyTorch" column is flagged as the
  pre-epic basic backend.
- Repointed the two `epic #9` xfail reasons to concrete issues (#30 legacy CLI save/load, #31
  legacy mixture M-step).

## Suggestions
- Re-export `AMICATorchNG` from `pyAMICA/__init__.py`; docstring `model_` notes both backends.

## Follow-up issues filed
- #26 adaptive PDF (NG), #27 multi-model partition matching, #30 legacy CLI save/load, #31 legacy
  mixture M-step, #32 remove parked `AMICATorchV2` + promote `backend="ng"` to default.

## Test plan
- [x] `uv run pytest pyAMICA/tests -m "not slow"` -> 61 passed, 2 xfailed
- [x] new wrapper tests -> 7 passed
- [x] `ruff check` / `ruff format` clean

### DIFF
--- a/.context/decisions/0001-torch-backend-natural-gradient-em.md
+++ b/.context/decisions/0001-torch-backend-natural-gradient-em.md
@@ -1,7 +1,7 @@
 # ADR 0001: Rewrite the PyTorch backend as a natural-gradient EM port, not Adam+autograd
 
-**Status:** proposed
-**Date:** 2026-07-02
+**Status:** accepted
+**Date:** 2026-07-02 (accepted 2026-07-03)
 **Owner:** Seyed Yahya Shirazi
 
 ## Context
@@ -36,6 +36,14 @@ statistics block-wise, and drop Adam/autograd for the parameter updates. The Num
   `AMICA` interface must be re-pointed at the new module.
 - **Obligation:** keep `validate_implementations.py` (real sample data + Fortran binary) green as
   the source of truth; no synthetic-data correctness tests.
+
+**Outcome (2026-07-03, issue #24 / PR #25):** implemented as `AMICATorchNG`
+(`torch_impl/amica_torch_ng.py`), wired into `AMICA(backend="ng")`. Single-model reaches Fortran
+parity (LL ~ -3.40, Hungarian component correlation ~0.997 with Newton positive-definite, 0
+fallbacks); the NumPy backend carries the same fixes. The root-cause fix was the natural-gradient
+A-update (it was transposed / multiplied on the wrong side); see
+`.context/issue-24/root_cause_Aupdate.py`. Multi-model M-step is bit-exact vs Fortran; full-fit
+partition matching is tracked in #27, adaptive-PDF in #26.
 
 ## Alternatives considered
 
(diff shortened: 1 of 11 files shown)


neuromechanist:
## Review (pr-review-toolkit, model=sonnet)

Three reviewers ran against commits `e27f717..bab24d1` (code, silent-failure, tests). All findings addressed in `80ad9f2`.

**code-reviewer:** No findings at confidence >= 80. Verified the MPS->CPU fallback on real MPS hardware with real sample data; confirmed explicit `device="mps"` still raises `AMICATorchNG`'s `ValueError`, CUDA is unaffected, and moving `import torch` to module top is safe. (Sub-threshold note: a `dtype="float64"` *string* bypasses both the wrapper and `AMICATorchNG`'s own guard, but that is pre-existing in the constructor, not a regression here, and violates the documented `dtype: torch.dtype` contract; left as-is.)

**silent-failure-hunter:**
- HIGH: the fallback notice was gated behind `verbose` -> now emitted via `logger.warning` unconditionally (matches `amica_torch.py`/`utils.py` device-caveat convention), with the `print` kept as a verbose-only nicety.
- MEDIUM: duplicated `torch.float64` default literal -> now derived from `inspect.signature(AMICATorchNG)` so the fallback can't silently stop firing if the default changes.
- MEDIUM: public docstring didn't mention the fallback -> added to the `device` parameter docs.
- Confirmed the fallback scoping (auto-only, explicit-mps still raises) is correct.

**pr-test-analyzer:**
- Added `test_ng_explicit_mps_float64_raises` (the negative complement of the fix; no MPS hardware needed).
- Added `test_ng_mps_float32_escape_hatch` (`skipif` no MPS) for the documented workaround.
- The fallback test now asserts the warning fires even with `verbose=False`.

Suite: 63 passed, 2 xfailed (0 failures); `ruff check`/`format` clean. No CI is configured on the repo.

PR #38, opened 2026-07-04
Add full-fit NumPy<->NG parity test (#37)

### LINKED ISSUE #37: Lock full-fit numerical parity between AMICA_NumPy and AMICATorchNG
We are keeping both the NumPy reference (`AMICA_NumPy`) and the PyTorch `AMICATorchNG` backend. `AMICATorchNG` was ported line-by-line from the NumPy `_get_block_updates`/`_update_parameters`, and the suite already proves per-block numerical parity to ~1e-8 (`test_sufficient_stats_match_numpy_reference`, `test_newton_stats_match_numpy_reference`, `test_newton_mstep_matches_numpy_reference`, `test_multimodel_sufficient_stats_match_numpy_reference`).

What is missing is an explicit **full-fit** head-to-head test: today each backend is validated only against the Fortran binary, not against each other end-to-end.

## Scope
- Add a `@pytest.mark.slow` real-data test that fits `AMICA_NumPy` and `AMICATorchNG` independently with the same single-model config and asserts they converge to the same solution: final LL within tolerance, and Hungarian-matched total-spatial-filter (`unmixing @ sphere`) correlation near 1.0.
- Thresholds set empirically from a measured run (not guessed) to avoid flakiness.

## Note on "identical"
Per-step update math is identical to ~1e-8 (already tested). A full multi-iteration fit is NOT bit-identical across NumPy and PyTorch (float reduction order / BLAS vs ATen), which is expected; the bar is "same converged solution", not bit-identical trajectories.

neuromechanist:
## Summary
Locks the numerical relationship between the two kept implementations (`AMICA_NumPy` and `AMICATorchNG`). We keep both — NumPy is NG's fast, readable oracle (the per-block parity tests copy params between them) and the CLI backend — so this adds the missing **full-fit** head-to-head check.

## What it does
`test_full_fit_parity_numpy_vs_ng` (slow) fits both backends independently on the real sample EEG data with a matched single-model config, then asserts they converge to the same solution:
- Hungarian-matched total-spatial-filter (`unmixing @ sphere`) correlation, bijective 32/32, `min > 0.99`.
- Same per-sample-per-channel LL (`|diff| < 1e-2`).

## Measured (thresholds set from data, not guessed)
- Hungarian-matched correlation: **min = mean = 1.000000**, 32/32 bijective.
- LL: NumPy reports the *un-normalized* total (−3,322,634); normalized by `n_samples * n_channels` that is **−3.40389**, matching NG's **−3.403887** to ~1e-5.

This complements the existing per-block sufficient-stat parity tests (which pin the update math to 1e-8) with an end-to-end convergence check. A full fit is intentionally not asserted bit-identical (float reduction order / BLAS vs ATen differ across NumPy and PyTorch); the bar is same converged solution.

## Related
- Closes #37. Follows the "keep both backends, lock parity" decision after #32.
- Side observation (not fixed here): NumPy `self.ll` is the un-normalized total while NG `ll_history` is normalized — a reporting inconsistency worth a small follow-up if we want them to print the same number.

## Testing
`test_full_fit_parity_numpy_vs_ng` passes (150s, two full 150-iter fits). ruff clean.

### DIFF
--- a/pyAMICA/tests/torch_tests/test_ng_backend.py
+++ b/pyAMICA/tests/torch_tests/test_ng_backend.py
@@ -731,6 +731,11 @@ def test_full_fit_parity_numpy_vs_ng(tmp_path):
         seed=SEED,
         max_decs=5,
         max_iter=150,
+        # AMICA_NumPy defaults do_opt_block=True, which auto-retunes block_size
+        # from data-timing at fit(), overriding the shared block_size=512 (NG
+        # has no such auto-tune). Disable it so both backends genuinely run the
+        # matched block size.
+        do_opt_block=False,
         outdir=str(tmp_path / "np_out"),
         **cfg,
     )
@@ -746,10 +751,12 @@ def test_full_fit_parity_numpy_vs_ng(tmp_path):
     a = filt_np / np.linalg.norm(filt_np, axis=1, keepdims=True)
     b = filt_ng / np.linalg.norm(filt_ng, axis=1, keepdims=True)
     corr = np.abs(a @ b.T)
+    # linear_sum_assignment on a square matrix returns a full permutation, so
+    # `matched` is already a bijective NumPy<->NG pairing; the correlation gate
+    # below is the real check.
     row, col = linear_sum_assignment(-corr)
     matched = corr[row, col]
 
-    assert len(np.unique(col)) == NW, "NumPy<->NG component matching is not bijective"
     assert matched.min() > 0.99, (
         f"min NumPy<->NG matched filter correlation {matched.min():.4f} <= 0.99"
     )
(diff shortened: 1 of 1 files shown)


neuromechanist:
## Review (code-reviewer, Sonnet) + resolutions

Ran an independent reviewer + two live runs. Findings addressed in the follow-up commit:

- **Configs weren't block-matched** (Important): `AMICA_NumPy` defaults `do_opt_block=True`, so `fit()` auto-retunes `block_size` from data timing (to 128 on this machine), overriding the shared `block_size=512`; NG has no auto-tune. Fixed by passing `do_opt_block=False` so both backends genuinely run at 512 (and NumPy's block size is no longer machine-dependent). Re-ran: still passes.
- **Tautological bijective assert** (Minor): `linear_sum_assignment` on a square matrix always returns a full permutation, so `len(np.unique(col)) == NW` couldn't fail. Removed it (replaced with a clarifying comment); the `matched.min() > 0.99` correlation gate is the real check.

Reviewer confirmed the compared quantities are apples-to-apples (`get_weights() @ sphere` vs `get_unmixing_matrix() @ sphere`, both `W.T` per the #24 convention) and the LL normalization is correct (NumPy total / (n_samples*n_channels) == NG's normalized LL). Real data only, correctly marked slow + skipif.

Issue #39, opened 2026-07-04
AMICA_NumPy long fit hits non-finite likelihood (NaN) ~iter 687 on sample data


neuromechanist:
Surfaced while fixing #30. With the CLI save/load format fixed, `test_sample_data_cli` (the full 2000-iter CLI run on the sample data with `sample_params.json`: `do_newton=true`, `newt_start=50`) now gets past the format step but the NumPy backend diverges to a non-finite log-likelihood:

```
Non-finite likelihood (NaN/-inf) encountered
Optimization finished after 687 iterations
```

The fit stops early at ~iter 687 and the degraded result no longer Hungarian-matches the Fortran reference (correlation < 0.8), so the integration test fails on `AssertionError: Some components don't match original results`.

This is distinct from #30 (format) and from the parity work: at ~150 iterations the NumPy backend is stable and matches Fortran (`test_sample_data_numpy_vs_fortran` and the #37 NumPy<->NG parity test both pass at 150 iters, ~0.997/1.0 correlation). The instability appears only in the longer run.

## Scope
- Reproduce and localize the NaN (likely an unclipped gradient / zero-denominator / Newton-step blow-up in the long run; see `.context/scratch_history.md` NaN notes).
- The torch `AMICATorchNG` backend runs the same update math without this blow-up on the sample data, so the NG path is a useful diff reference.
- Once fixed, un-xfail `test_sample_data_cli`.

Note: `AMICATorchNG` (the default backend) is unaffected; this is specific to the legacy NumPy long fit.

Issue #15, opened 2026-07-02
Test coverage gaps: AMICA.save/load and plot_components


neuromechanist:
Surfaced during Phase 1 review (epic #9, PR #14).

- `AMICA.save`/`AMICA.load` (pyAMICA/amica.py:244-297) do a real torch.save/load state_dict roundtrip but have zero test coverage. The old `AMICATorch.save/load` test was deleted (those methods no longer exist); coverage moved to the wrapper and was never re-added.
- `amica_viz.plot_components` (pyAMICA/amica_viz.py:84) has a current signature `(results_dir, data=None, ...) -> None` and is fully untested since the stale `test_visualization` was deleted.

Add real (non-mock) tests: save/load a fitted model to tmp_path and assert roundtrip equality of W/A; render plot_components from a real saved results directory and assert files are written.

neuromechanist:
The **save/load** half of this issue is now covered by #44 (merged): real fit -> save -> load -> transform round-trip plus unfitted/degenerate/non-finite/snapshot guard tests (`test_amica_ng_wrapper.py`, `test_ng_backend.py::test_state_dict_*`).

Remaining under this issue: **`amica_viz.plot_components`** test coverage. Keeping open for that.

neuromechanist:
Done via #65 (and #36). AMICA.save/load roundtrip is covered by test_amica_ng_wrapper.py (save-requires-fit, save/load roundtrip asserting W/A equality, version guard, missing-key guard, degenerate-refusal) added under #36. plot_components builds an in-memory figure and never writes a file, so the 'assert files are written' premise does not apply; #65 covers its previously-untested activation branch (data provided) instead. amica.py coverage 86% -> 96%.

Issue #27, opened 2026-07-03
Multi-model NG partition matching vs Fortran (>0.95 cross-corr)


neuromechanist:
Follow-up from #24 (parity PR #25).

The multi-model (`n_models>1`) M-step in `AMICATorchNG` is proven bit-exact vs Fortran: one seeded 2-model step matches per-model unmixing to ~1e-15 and `gm` exactly, the sphere matches to ~1e-13, and NG<->NumPy multi-model sufficient-stat parity holds to 1e-8 (suite test `test_multimodel_sufficient_stats_match_numpy_reference`).

However, a full 2-model fit reaches a comparable LL (NG -3.355 vs Fortran -3.345) but only ~0.77 Hungarian cross-correlation, because multi-model AMICA has many near-degenerate partitions (unlike single-model's unique solution). NG is self-consistent (cross-corr 1.0 across block sizes), so this is intrinsic partition ambiguity, not a code defect.

Open question (may be unachievable): can NG be made to converge to Fortran's exact partition (>0.95 cross-corr)? Candidate levers:
- Exact fit-loop-schedule parity with Fortran (lrate ratchet / numdecs / Newton ramp) for multi-model.
- Seeded, sphere-matched runs to reduce transient divergence.

Note: this is speculative; the algorithmic correctness is already validated. Document outcome even if it stays partition-limited.

neuromechanist:
PR #25 review (code-reviewer) surfaced a concrete, fixable contributor to the multi-model gap that is NOT partition ambiguity: **the per-model bias `c` update is omitted.**

- Fortran's `update_c` defaults on: `c[i,h] = dc_numer[i,h]/dc_denom[i,h]`, `dc_numer[i,h] = sum_t v_h(t)*data(i,t)`, `dc_denom[i,h] = sum_t v_h(t)` (amica17.f90:1423-1429, 1899-1901). It is a per-model, per-channel (data-space) responsibility-weighted mean, subtracted before W.
- For `n_models=1`, `v_h ≡ 1`, so `dc_numer = sum(data) = 0` on mean-removed data -> c stays 0 -> single-model parity exact (unaffected).
- For `n_models>1`, `v_h` is non-uniform, so Fortran moves c per model; both Python backends freeze c at 0. This diverges the multi-model trajectory.
[...]

PR #40, opened 2026-07-04
Write CLI results in Fortran binary format (#30)

### LINKED ISSUE #30: Legacy CLI save/load format mismatch (.npy vs Fortran binary)

neuromechanist:
## Summary
The legacy NumPy CLI wrote results as `.npy`, but `loadmodout` (and the tests/comparison path) expect raw Fortran-format binaries with no extension, so the round-trip raised `FileNotFoundError` for `W`. This converges pyAMICA on the single Fortran-binary format.

## Changes
- **`AMICA._write_results`** now writes raw little-endian float64 (int32 `comp_list`) files (`gm, A, W, S, mean, c, alpha, mu, sbeta, rho, comp_list, LL`) in `loadmodout`'s byte layout, so pyAMICA output is directly comparable to and loadable by the same reader as the Fortran `amicaout`. `comp_list` is written 1-based (Fortran convention).
- **`load_results`** now reads that raw-binary format back into AMICA's internal array shapes for the `amica_viz` helpers (the `compressed` flag is a documented no-op; the format is never `.npz`).
- Removed the dead `amica_data.save_results` (`.npy`/`.npz` writer, no callers) to finish converging on one format.
- Fixed a pre-existing broken import in `amica_viz.plot_pdf_fits` (`from amica_pdf` → `from .amica_pdf`).
- Added a fast, non-slow smoke test `test_cli_output_format_roundtrip` covering the format round-trip (`loadmodout` + `load_results` shapes) and all three viz helpers on real data (also viz coverage toward #15).

## Verification
- Round-trip: fit → `_write_results` → `loadmodout` Hungarian-matches the Fortran reference `W` at **min corr 0.95 / 32-of-32 bijective** (150-iter fit; the full run is higher).
- New smoke test passes (2.4s); non-slow suite green (33 passed, 1 xfailed).

## Honest status on `test_sample_data_cli`
The #30 format bug is fixed, but I kept `test_sample_data_cli` (the full 2000-iter CLI-vs-Fortran integration test) `xfail`: with the format fixed it now gets past the load step, but the **NumPy backend diverges to a non-finite likelihood ~iter 687** on the sample data and stops early, so the degraded result no longer matches Fortran. That is a **separate long-fit numerical-stability bug**, filed as **#39** (the torch `AMICATorchNG` default backend is unaffected; the NumPy backend is stable and matches Fortran at ~150 iters per the #37 parity test). The xfail reason now points at #39, and the format contract itself is covered by the new smoke test.

## Related
- Closes #30.
- Files #39 (NumPy long-fit NaN instability).

### DIFF
--- a/pyAMICA/amica_data.py
+++ b/pyAMICA/amica_data.py
@@ -9,7 +9,7 @@ algorithm. It handles:
    - Mean removal
    - Sphering (whitening) using PCA
    - Dimensionality reduction
-3. Result saving and loading with optional compression
+3. Loading saved results (raw Fortran binary format; see load_results)
 
 The preprocessing steps are crucial for AMICA's performance:
 - Mean removal centers the data, removing DC offset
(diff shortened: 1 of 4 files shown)


neuromechanist:
## Review (3 agents, Sonnet) + resolutions

Findings addressed in b20ba66:

### Addressed
- **No value-level round-trip assertion** (test, sev 7): the smoke test checked only shapes, so a transpose/axis-order bug in the W byte layout would pass. Added value asserts: loaded `W`/`A`/`alpha`/`comp_list` == the in-memory arrays. (The comment reviewer independently confirmed the round-trip is byte-exact for W/A.)
- **No real CLI-path test** (test, sev 6): added `test_cli_subprocess_output_loadable`, which runs the actual `amica_cli` entrypoint (the thing that raised `FileNotFoundError`) on a short stable config and asserts `loadmodout` reads it. Fast (~5s), non-slow.
- **No final write after fit** (silent-failure / test): `_write_results` only fired on `writestep` boundaries, so a run whose last iteration isn't a `writestep` multiple never persisted the converged result. `AMICA.fit` now does an unconditional final write, **finite-guarded** so a #39 NaN run doesn't overwrite the last good output. The smoke test now relies on this path.
- **Multi-model layout claim false** (code-review, Important): the C-order dump is byte-identical to Fortran only for `num_models==1` (the CLI/sample case). Scoped the docstring/comments honestly; multi-model output is self-consistent (loadmodout round-trips) but not byte-identical to Fortran (noted, relates to #27).
- **`load_results` silently accepts incomplete dirs** (silent-failure): now raises a clear `FileNotFoundError` listing the missing core files instead of a cryptic downstream `KeyError` in a viz helper.
- **Stale `compressed` docstrings** (silent-failure): updated across the viz helpers to note it's a legacy no-op.

### Skipped (with rationale)
- **`_write_results` write atomicity** (silent-failure): a crash mid-write leaves a partial dir. Pre-existing (the old .npy writer had the same), and temp-file+rename is a broader change; the new `load_results` incomplete-dir check now at least fails loudly on such a dir. Worth a separate follow-up.
- **`num_comps != nw*num_models` (PCA reduction) shape derivation** (both reviewers, low): only affects the PCA-reduced path, which is pre-existing and non-functional (`data_dim` is never updated after reduction, unrelated to this diff). Out of #30 scope.
- **`test_sample_data_cli` has no `raises=`**: it can't distinguish the #39 NaN from a writer regression, but the new value asserts + CLI subprocess test now guard writer regressions directly, so the slow xfail doesn't need to.

Non-slow suite after fixes: 34 passed, 1 xfailed (the unrelated under-convergence test).

Issue #41, opened 2026-07-05
AMICA_NumPy long fit drifts/degrades past ~150 iters (LL decreases, component match drops)


neuromechanist:
Surfaced while resolving #39. With a fixed stable seed (seed=0) the NumPy backend no longer NaNs on the full 2000-iter sample config (the #39 restart mechanism + pinned seed), but the solution **degrades over the long run**:

| iters | normalized LL | Hungarian W-match vs Fortran (min / mean, n>0.8) |
|------:|--------------:|--------------------------------------------------|
| 150   | -3.4043       | 0.921 / 0.970, 32/32                              |
| 2000  | -3.4093       | some < 0.8 (test fails)                           |

The LL *decreases* between iter ~150 and 2000 (worse solution with more iterations) and several component correlations drop below the 0.8 gate. At ~150 iters the match is excellent (32/32), so the backend finds the right solution and then drifts away from it.

The natural-gradient A-update is a gradient step (with lrate), not strict EM, so it can overshoot/oscillate. Fortran runs the same 2000-iter config and stays converged (the reference `amicaout`), so this is a **convergence-schedule / stability parity gap** (lrate ratchet / numdecs / Newton ramp), distinct from:
- #39 (the NaN, now handled by restart + seed), and
- #27 (multi-model partition ambiguity).

## Scope
- Reproduce and localize the drift (does `max_decs`/`min_dll`/lrate-reduction fire? does the LL oscillate?).
- Align the long-run lrate/Newton schedule with Fortran so a converged solution is not degraded by over-iterating.
- Then un-xfail `test_sample_data_cli` (the full 2000-iter CLI-vs-Fortran integration test).

Repro: fit `AMICA_NumPy` with the sample `sample_params.json` config at seed=0, compare loadmodout-W Hungarian match at 150 vs 2000 iters.

neuromechanist:
Partially resolved by PR #43.

[...]

PR #42, opened 2026-07-05
Add restart-on-NaN mechanism to NumPy backend (#39)

### LINKED ISSUE #39: AMICA_NumPy long fit hits non-finite likelihood (NaN) ~iter 687 on sample data

neuromechanist:
## Summary
Ports Fortran's NaN-recovery to the NumPy backend. The observed failure (CLI diverging to a non-finite likelihood ~iter 687) is a **stochastic, init-dependent** instability: with a fixed seed the fit is stable, and NG does the identical unfloored exact-EM `mu`/`beta` update, so it is a shared algorithm property, not a NumPy-specific bug. Fortran's only mitigation is a restart-on-NaN, which the NumPy backend lacked entirely.

## Changes
- **Restart-on-NaN** (Fortran `amica17.f90:1027-1060`): if the LL goes non-finite at `iter <= restartiter` (default 10), reinitialize from a fresh RNG draw and start over, up to `maxrestarts` (default 3); a later NaN stops the fit (Fortran exits too). New `restartiter`/`maxrestarts`/`numrestarts` params and a `_reinitialize_for_restart` helper.
- **Convergence-guard fix**: `_check_convergence` guarded its `ll[-2]` comparison on `self.iter`, which keeps counting across restarts; guard on `len(self.ll)` instead so a post-restart first iteration doesn't `IndexError`.
- **Tests**: fault-injection tests for the restart-recover and give-up-after-maxrestarts paths (real data; a subclass injects a non-finite LL to exercise the error path deterministically). Pinned `--seed 0` in `test_sample_data_cli` so the integration test doesn't depend on a random draw.

## Verification
- Restart tests pass; non-slow suite 36 passed, 1 xfailed.
- The full 2000-iter CLI run at seed=0 now **completes without NaN** (finite LL ~-3.41), where a random seed previously diverged.

## Honest status on the CLI integration test
`test_sample_data_cli` stays **xfail** — this PR is progress, not a pass. The NaN is handled (seed pin + restart => all 2000 iters, no early stop), but a **separate long-run drift** remains: at ~150 iters seed=0 matches Fortran 32/32 (min corr 0.92), yet by 2000 iters the LL degrades (-3.404 -> -3.409) and some component correlations drop below the 0.8 gate. Fortran stays converged over the same run, so this is a convergence-schedule/stability parity gap, filed as **#41**. The xfail reason points at #41.

## Related
- Closes #39 (restart-on-NaN mechanism; the NaN is now handled).
- Files #41 (NumPy long-run drift/degradation).

### DIFF
--- a/pyAMICA/amica_cli.py
+++ b/pyAMICA/amica_cli.py
@@ -27,6 +27,7 @@ And many others as documented in the AMICA class.
 import argparse
 import json
 import logging
+import sys
 from pathlib import Path
 from typing import Dict, Any
 
@@ -47,31 +48,20 @@ def parse_args() -> argparse.Namespace:
         - seed: Random seed for reproducibility (optional)
         - verbose: Flag for detailed per-line progress output (disables tqdm progress bar)
     """
-    parser = argparse.ArgumentParser(
-        description='AMICA: Adaptive Mixture ICA'
-    )
+    parser = argparse.ArgumentParser(description="AMICA: Adaptive Mixture ICA")
 
     # Required arguments
-    parser.add_argument(
-        'paramfile',
-        help='Parameter file in JSON format'
-    )
+    parser.add_argument("paramfile", help="Parameter file in JSON format")
 
     # Optional arguments
     parser.add_argument(
-        '--outdir',
-        help='Output directory (default: output)',
-        default='output'
+        "--outdir", help="Output directory (default: output)", default="output"
     )
+    parser.add_argument("--seed", help="Random seed", type=int)
     parser.add_argument(
-        '--seed',
-        help='Random seed',
-        type=int
-    )
-    parser.add_argument(
-        '--verbose',
-        help='Verbose output with detailed per-line progress (disables tqdm progress bar)',
-        action='store_true'
+        "--verbose",
+        help="Verbose output with detailed per-line progress (disables tqdm progress bar)",
+        action="store_true",
     )
 
     return parser.parse_args()
@@ -114,16 +104,11 @@ def load_params(paramfile: str, default_paramfile: str = None) -> Dict[str, Any]
         params.update(user_params)
 
     # Required parameters
-    required = {
-        'files',
-        'data_dim',
-        'field_dim'
-    }
+    required = {"files", "data_dim", "field_dim"}
 
     missing = required - set(params.keys())
     if missing:
-        raise ValueError(
-            f"Missing required parameters: {', '.join(missing)}")
+        raise ValueError(f"Missing required parameters: {', '.join(missing)}")
 
     return params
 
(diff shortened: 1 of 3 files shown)


neuromechanist:
## Review (code-reviewer + silent-failure-hunter, Sonnet) + resolutions

Findings addressed in the latest commit:

### Addressed
- **CRITICAL (parity): restart cold-reset too much.** `_reinitialize_for_restart` was nulling `mu/alpha/beta/rho/gm/c`, but Fortran's restart (`amica17.f90:1032-1053`) redraws **only `A`** (recomputing `comp_list`/`W`) and keeps the last-successful mixture params, skipping the M-step that iteration. Now redraws only `A`.
- **CRITICAL/HIGH (silent failure): success reported on divergence.** After `maxrestarts` exhaustion the fit ends on a NaN and writes nothing, yet the CLI logged "Results saved" and exited 0, and a library caller had no signal. `AMICA.fit` now sets `self.converged`/`self.stop_reason`; the CLI logs an ERROR and exits non-zero (no false "saved"); restart logs at WARNING, terminal divergence at ERROR.
- **`numrestarts` not reset on refit** → reset at the start of each fit (per-fit budget).
- **Window/count fidelity** → use Fortran's absolute `iter <= restartiter` window (dropped the more-lenient per-restart offset); comment no longer claims byte-for-byte parity on the restart count.
- **Tests** now assert `converged` (True on recover, False on give-up) and that the restart is announced in the run log (guards against a silent-restart regression). The `_check_convergence` `len(self.ll)` guard fix is retained.

### Verified / not changed
- Restart loop is bounded (numrestarts cap), `continue` correctly skips convergence/reject/share/write, post-restart E-step uses the redrawn `A`, and the finite-guarded final write never persists NaNs (confirmed by `test_restart_gives_up_after_maxrestarts`).
- Newton state and `comp_list`/`W` are rebuilt by `_initialize_parameters`; RNG is preserved so each restart is a genuinely new basin.

Non-slow suite: 36 passed, 1 xfailed. The full CLI integration test remains xfail on the separate #41 long-run drift (as documented).

PR #43, opened 2026-07-05
Ratchet the learning rate on LL decrease (#41)

### LINKED ISSUE #41: AMICA_NumPy long fit drifts/degrades past ~150 iters (LL decreases, component match drops)

neuromechanist:
## Summary
The NumPy long fit degraded past ~150 iters (LL decreasing, component match dropping) because pyAMICA's convergence handling diverged from Fortran in two ways:

1. **`numdecs`/`numincs` never accumulated.** `_check_convergence` incremented them locally but never returned them, so the caller's counters stayed at their initial values — the decrease handling effectively reset every iteration.
2. **On `numdecs >= max_decs` pyAMICA stopped**, whereas Fortran (`amica17.f90:1062-1108`) lowers the learning-rate **ceiling** (`lrate0`, plus `newtrate`/`rholrate0` under Newton), resets `numdecs`, and **continues**. Without the ceiling ratchet, `lrate` kept ramping back up (it's raised each iter in `_update_parameters`) and the fit oscillated/drifted.

## Changes
- Rewrote `_check_convergence` to match Fortran: thread `numdecs`/`numincs` back into the loop, ratchet `lrate0`/`newtrate`/`rholrate0` down at `max_decs` and **continue** (only a lrate/gradient floor terminates on a decrease), stop after `maxincs` consecutive small increases (new `maxincs` param, Fortran default 5), and reset `numdecs` when Newton starts.

## Effect
- seed=0: the LL now **improves** over the long run — **-3.399** at 2000 iters, vs the pre-fix -3.409 and the ~150-iter -3.404 (and better than Fortran's -3.402). No more degradation.
- The 150-iter `test_sample_data_numpy_vs_fortran` (mean-corr > 0.9) still passes — no parity regression.
- Non-slow suite: 36 passed, 1 xfailed.

## Honest status on `test_sample_data_cli`
Still **xfail**. The #41 **LL-degradation drift is fixed**, but the test's strict *all-32-components* > 0.8 gate still fails (~27/32): at full convergence pyAMICA reaches a comparable/slightly-better-LL but **different-partition** solution than the Fortran reference on a few ill-determined components — a solution-basin difference (the single-model tail of #27's partition theme), not the drift. The project's other parity test uses the more appropriate `mean > 0.9` bar and passes. The xfail reason documents this.

## Related
- Addresses #41 (LL-degradation drift fixed; the numdecs accumulation bug and the missing lrate ratchet).

### DIFF
--- a/pyAMICA/pyAMICA.py
+++ b/pyAMICA/pyAMICA.py
@@ -205,6 +205,9 @@ class AMICA:
         self.pcadb = params.get("pcadb")
         self.writestep = params.get("writestep", 100)
         self.max_decs = params.get("max_decs", 5)
+        # Consecutive small-increase iterations tolerated before stopping
+        # (Fortran maxincs, amica17.f90:1087).
+        self.maxincs = params.get("maxincs", 5)
         # Restart-on-NaN (Fortran amica17.f90:1027-1060): if the LL goes
         # non-finite at iter <= restartiter, reinitialize and start over, up to
         # maxrestarts times; a later NaN stops the fit (Fortran exits too).
@@ -1114,11 +1117,19 @@ class AMICA:
                         if self.verbose or not self.use_tqdm:
                             self.logger.info(detailed_log)
 
-                # Check convergence
-                converged, reason = self._check_convergence(numdecs, numincs)
+                # Check convergence (threads numdecs/numincs back so they
+                # accumulate across iterations, and ratchets the lrate ceiling).
+                converged, reason, numdecs, numincs = self._check_convergence(
+                    numdecs, numincs
+                )
                 if converged:
                     convergence_reason = reason
                     break
+
+                # Reset the decrease counter when Newton turns on (Fortran
+                # amica17.f90:1105-1108).
+                if self.do_newton and iter == self.newt_start:
+                    numdecs = 0
 
                 # Reject outliers if requested
                 if (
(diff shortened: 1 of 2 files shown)


neuromechanist:
## Review (code-reviewer, Sonnet) + resolutions

The reviewer confirmed the ratchet logic matches Fortran (`amica17.f90:1062-1108`) line-by-line -- the decrease-branch stop condition, the `lrate0`/`newtrate`/`rholrate0` reductions and their `iter>newt_start`/`do_newton` guards, the `numdecs` reset, the `numincs`/`maxincs` handling (default 5, matching the Fortran header), the counter threading, and the `newt_start` reset placement. No hang/regression risk; the `lrate0` ratchet is confirmed to actually affect the lrate ramp ceiling.

Addressed in the follow-up commit:
- **Important (parity gap for `use_grad_norm=False`):** the decrease-branch gradient floor was gated by `use_grad_norm`, but Fortran's `ndtmpsum <= min_nd` decrease-stop is unconditional (the flag only gates the *final* grad-norm stop). Since `self.nd` was only populated when `use_grad_norm` was set, the floor silently dropped for `use_grad_norm=False`. Now the gradient norm is computed **every iteration** (matching Fortran), and the decrease-stop floor applies unconditionally; `use_grad_norm` still gates only the separate final stop.
- **Coverage (non-blocking):** added a direct unit test for the ratchet (`test_check_convergence_ratchets_lrate_on_decrease`) so the counter-threading/ceiling-ratchet behavior is regression-guarded without relying on the slow integration test.

Non-slow suite: 37 passed, 1 xfailed (the CLI integration test, on the residual partition-basin difference documented in its xfail reason).

PR #68, opened 2026-07-07
feat: component sharing (share_comps) in AMICATorchNG

### LINKED ISSUE #60: Implement component sharing (share_comps) in AMICATorchNG
Last unimplemented Fortran feature. In `amica15.f90`, `comp_list` can share components across models via the `share_comps` reassignment on the `share_start`/`share_int` schedule (plus `load_comp_list`, `comp_used`). In `AMICATorchNG`, `comp_list` is only the trivial per-model block layout (`comp_list[:, h] = arange(h*n, (h+1)*n)`); the reassignment is absent. NumPy has partial plumbing (`params.json` `share_comps`/`share_start`/`share_int`, `load_comp_list`) but not the dynamic reassignment either.

## Tasks
- [ ] Read the `share_comps` path in `amica15.f90` (comp_used bookkeeping, reassignment cadence, comp_thresh).
- [ ] Port `share_comps` reassignment + `load_comp_list` init to `AMICATorchNG`.
- [ ] Reconcile the NumPy plumbing so both backends agree.
- [ ] Validate on real sample data vs Fortran (partition/component agreement); guard no-op for `n_models=1` so single-model #24 parity stays bit-exact.

## Priority
Low per `.context/feature_parity.md` (`share_comps` is niche), but it is the only remaining feature-parity gap.

neuromechanist:
Closes #60. Ports the last unimplemented Fortran feature: multi-model component sharing.

## What it does
On the `share_start`/`share_iter` schedule, components that are near-collinear **across different models** (cosine angle of their de-sphered mixing columns above `comp_thresh`) are merged into one shared mixing column and one shared density (Fortran `identify_shared_comps`, amica15.f90:1898). An A-freeze holds the mixing update for ~6 iterations after each merge so densities can settle.

## Design (mirrors the #26 adaptive-PDF precedent)
- **No bit-exact oracle.** The reference's similarity metric `Spinv2` is **never declared/initialized** in `amica15.f90` (which has `implicit none`, so it would not even compile as-is) -- its reassignment never actually fires, exactly like the dead `do_choose_pdfs` switch (#26). So this implements the *intended* algorithm (cosine similarity of de-sphered scalp maps) and validates by real-data behavior, not byte parity.
- **Off by default**, and a no-op for `n_models=1` -- single-model (#24) and default multi-model (#27) parity stay **byte-for-byte** (full torch suite green: 95 passed).
- The M-step already sums sufficient stats through `comp_list` (index_add), so shared densities update jointly for free. The **A-update was refactored** to accumulate shared columns the same way (`dAk` index_add) -- byte-identical to the per-model update when unshared.
- **Merged-away columns are frozen** in the density M-step (they get no stats, so `0/0 -> NaN` otherwise; Fortran tolerates this behind its `comp_used` mask, we keep them finite so save/the degenerate-fit contract aren't tripped).

## Tests (`tests/torch_tests/test_ng_sharing.py`, 6)
- single-model byte-identical with sharing on vs off
- controlled single collinear-pair merge
- the 'no double-share within a model' guard (all-collinear 2x2 -> 2 shared comps, not 1)
- default multi-model leaves comps unshared
- full 2-model share fit stays finite and shares
- save/load round-trips the share config + merged comp_list

No new persisted state (comp_list already persisted; `comp_used` is derived). Also gitignores `.env`.

### DIFF
--- a/.context/feature_parity.md
+++ b/.context/feature_parity.md
@@ -8,8 +8,7 @@ and CLI. Living status is in `../AGENTS.md` (Known Issues); ADRs are in `decisio
 **Headline:** single-model parity is done (LL ~ -3.40 vs Fortran -3.4018; Hungarian component
 correlation ~0.997 with Newton positive-definite, 0 fallbacks). Adaptive PDF (#26), multi-model
 (#27, distributional equivalence), best-iterate (#51), and the degenerate-fit contract (#50) are
-done. The one unimplemented Fortran feature is component sharing; the one unverified success
-criterion is runtime.
+done, as is component sharing (#60). The one unverified success criterion is runtime.
 
 ## Core AMICA features
 
@@ -45,7 +44,7 @@ criterion is runtime.
 |---------|---------|--------------|--------|-------|
 | Outlier rejection | yes | yes | Complete | `do_reject` |
 | Adaptive PDF selection | yes | yes | Complete | 5 `pdftype` families + ext-Infomax switcher (#26) |
-| Component sharing | yes | **no** | **Missing** | `share_comps` reassignment not ported (torch `comp_list` is trivial per-model layout) |
+| Component sharing | yes | yes | Complete | `share_comps` multi-model merge ported (#60); off by default, behavior-validated (no bit-exact oracle) |
 | History tracking | yes | yes | Complete | `ll_history`, grad norms |
 
 ## Input/output
@@ -91,7 +90,7 @@ indistinguishable from Fortran's run-to-run distribution (#27, `.context/issue-2
 
 - [ ] **Performance benchmark** vs the Fortran binary (CPU/CUDA/MPS); verify or revise the 2-3x
       runtime criterion.
-- [ ] **Component sharing** (`share_comps`, `share_start`/`share_int`, `load_comp_list`,
-      `comp_used`) ported to `AMICATorchNG` and validated on real data.
+- [x] **Component sharing** (`share_comps`, `share_start`/`share_iter`, `comp_thresh`) ported to
+      `AMICATorchNG` (#60), off by default, behavior-validated on real data.
 - [ ] **Test hardening:** single-channel / single-sample edge cases, numerical-stability
       regression tests, and `save`/`load` + `plot_components` coverage (#15).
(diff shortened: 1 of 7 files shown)


neuromechanist:
## Review complete (4 reviewers, Sonnet) + all findings addressed

Ran the pr-review-toolkit reviewers (code, tests, silent-failure, comments) in parallel. The default `share_comps=False` path was verified **byte-identical** (single-model #24 exact, multi-model #27 preserved); the real bugs were all in the *merged* path and are now fixed.

### Critical
- **A-update was a raw sum, not Fortran's `gm`-weighted average** (`dAk/zeta`, amica15.f90:1730-1743). The raw sum over-steps by the contributor count and destabilized the fit (LL dropped after a merge and never recovered). Ported the weighted average; single-model `gm=[1]` cancels exactly so #24 stays byte-identical.
- **`keep_best` (#51) silently reverted merges** by restoring `comp_list` from a pre-merge snapshot -> the returned model was unmerged under the default `keep_best=True`. Disabled `keep_best` under `share_comps` (+ warning), like `do_reject`; returns the last merged iterate (Fortran).
- **`_a_frozen` froze A permanently for `share_iter<=6`** and misaligned with the merge schedule. Anchored on `(itf-share_start)%share_iter` and require `share_iter>6`.

### Important
- **Stale `W` after a merge:** `_identify_shared_comps` ran after `W` was rebuilt, so the next E-step read a pre-merge `W` while indexing densities by the merged `comp_list`. Now rebuild `W` immediately after the merge (Fortran order).
- **Unguarded sphere inversion:** `inv(sphere)` crashes / returns garbage on average-referenced or rank-deficient EEG, and a NaN `t0` fell through `t0<thresh`. Added an inversion guard (clear error), reject `share_comps`+`pcakeep`/`pcadb` up front, and guard `t0` finiteness.

### Low
Newton-fallback counter no longer polluted during a freeze; `alpha` added to the non-finite check; `_spinv` reset per `fit()`.

### Comments/docs
`Spinv2` reframed as *declared-but-never-allocated* (undefined behavior / crash, not a benign no-op); `amica17`->`amica15:1786-1797` citation corrected; "after each merge"->periodic-window.

### Tests: 8 -> 19
Added constructor validation, the `_a_frozen` window schedule, a 3-model guard case, `comp_thresh=1.0`, singular-sphere-raises, **merge-survives-to-final-model**, and **unique-count-drops-without-LL-degradation**. Full torch suite green (108 passed, 5 skipped). No findings skipped.

PR #69, opened 2026-07-08
perf: drop dead dpdf + dedup |y|^rho in E-step (#63)

### LINKED ISSUE #63: Optimize AMICATorchNG per-iteration cost (pow hot spot + CPU threading)
Follow-up from the #59 benchmark. On the CPU float64 parity path, NG is ~3.2x slower than Fortran single-threaded and ~7.3x at 8 threads (Fortran even handicapped by Rosetta here). Two root causes, from a cProfile of a 20-iter fit (see #59):

## 1. `pow` is ~49% of runtime
2.55s of 5.15s is `Tensor.pow`, from the generalized-Gaussian `|z|^rho` (fractional exponent, single-threaded). It is evaluated separately in `_log_pdf_and_deriv` (core.py:94) and `_score` (core.py:159), so the density and its derivative each pay a fractional pow.

Proposals (all must stay bit-exact vs the Fortran `z0`/`fp` tests in `test_ng_pdf_families.py`):
- Compute `absz_pow = |z|**rho` once and derive the score from it (`fp ~ rho * absz_pow / z`) instead of a second fractional pow.
- Special-case `rho == 2` (square) and `rho == 1` (abs) to skip `pow` entirely for those PDF families.
- Consider caching `log|z|` where `|z|^rho = exp(rho*log|z|)` is reused across pdf/score/rho-update.

## 2. NG does not parallelize on CPU
Going 1 -> 8 torch threads moves per-iter only 219 -> 204 ms (~7%), while Fortran OpenMP scales 68 -> 28 ms (~2.4x). The per-block Python loop dispatches many small elementwise kernels; intra-op threading cannot amortize that.
- Investigate larger/vectorized block handling to give torch bigger ops to thread.
- GPU (CUDA) is NG's intended fast path (float64 parallelizes far better); MPS cannot do float64. Quantify on a CUDA host via the #59 harness.

## Acceptance
- Single-model #24 parity stays byte-for-byte; PDF-family bit-exactness preserved.
- Re-run `benchmarks/benchmark_runtime.py`; record the new ratio in this issue.

neuromechanist:
Part of #63. The clear, **bit-identical** win from the benchmark's pow hot spot.

## What
The forward pass computed the density derivative `dpdf` every block, but **every caller discards it** (the exact-EM M-step uses the score `fp`, not `dpdf`) -- so a `|y|^(rho-1)` power and **three exp()s** per block-model were dead work. Separately, `|y|^rho` was computed twice (density + the rho-update accumulator).

Added a lean `_log_pdf_only` that computes only `log_pdf` and returns `|y|^rho`; `_forward` threads it through to `_get_block_updates`, which reuses it instead of recomputing. `_log_pdf_and_deriv`/`_score` are untouched (still cover the pdf-family tests).

## Safety
**Bit-identical** -- single-model final LL matches the pre-change value to all 15 digits (verified by stashing the change and re-running), and the full torch suite is green (108 passed, 5 skipped). The parity tests (NumPy 1e-10, families 1e-12) all pass unchanged.

## Measured (real sample EEG, single thread)
| Metric | Before | After |
|--|--|--|
| pow time (20 it) | 2.55s | **1.30s** (-49%) |
| 30-iter wall | 255 ms/it | **166 ms/it** (-35%) |

## Not in this PR (need your call -- data in the PR thread)
- Larger `block_size` default (128->512): another ~-18%, but shifts results ~6e-6 (accumulation order).
- CPU threading: ~-16% at 4 threads, degrades beyond (workload-limited).
- GPU float32 fast path: the real GPU lever; needs a CUDA host to validate.

### DIFF
--- a/.context/issue-63/perf_findings.md
+++ b/.context/issue-63/perf_findings.md
@@ -37,32 +37,33 @@ A single iteration's sufficient stats are block-size-independent to ~1e-8 (test
 it changes (chaotic optimization compounds the 1e-8). All tests specify
 block_size explicitly, so the default change is safe (108 passed).
 
-## 3. GPU: CUDA float64 is a clean 2.14x (NVIDIA RTX 4090)
+## 3. GPU: CUDA float64 is a clean ~4.5x (NVIDIA RTX 4090)
 
-Real device numbers from `hallu` (RTX 4090), full data, 50 iters,
-`benchmarks/benchmark_gpu.py`:
+Real device numbers from `hallu` (RTX 4090), full data, 50 iters, **warmed**
+(first CUDA call pays context init + kernel compilation, ~2-3x inflation if not
+excluded), min of 4 repeats, `torch.set_num_threads(16)`:
 
 | device | dtype | ms/it | vs CPU-f64 | final LL |
 |---|---|---|---|---|
-| cpu | float64 | 281 | 1.00x | -3.42408 |
-| **cuda** | **float64** | **131** | **2.14x** | **-3.42408** (identical) |
-| cpu | float32 | 54 | 5.24x | **NaN** |
-| cuda | float32 | 35 | 8.05x | **NaN** |
+| cpu (16 thr) | float64 | 172.8 | 1.00x | -3.42408 |
+| **cuda** | **float64** | **38.5** | **4.5x** | **-3.42408** (identical) |
 
-**CUDA float64 is the safe GPU win**: 2.14x and numerically identical to CPU
-(-3.42408). The `AMICA` wrapper auto-selects CUDA when present.
+**CUDA float64 is the safe GPU win**: 4.5x over a 16-thread CPU (~7x over
+single-thread) and numerically identical to CPU (-3.42408 to the digit). The
+`AMICA` wrapper auto-selects CUDA when present. (Measurement caveat: a cold first
+CUDA call reads ~131 ms/it -- always warm up before timing GPU.)
 
 ## 4. float32 is precision-limited, NOT a free fast path
 
-float32 would be 5-8x faster, but on the full 30504-sample data it **collapses to
-NaN around iter 23**: it descends healthily (-3.51 -> -3.44 over 22 iters), then a
-mixture component's responsibility mass underflows in float32's ~7-digit range,
-the exact-EM `0/0` produces non-finite mu/beta/alpha, and the degenerate-fit
-contract (#50) stops it with a clean `nan_ll` (it fails LOUDLY, not silently). It
-does converge on smaller slices (4096 samples: -3.234, matching float64). So
-float32 is usable for small data / experimentation but not production on
-full-size recordings without float32-specific mixture flooring -- tracked as a
-follow-up.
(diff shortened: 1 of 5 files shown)


neuromechanist:
## #63 complete: measured on CPU + a real RTX 4090 (`hallu`)

Went beyond the pow-dedup to cover all three levers you asked for, with real GPU numbers.

### Landed here (float64, safe)
| Lever | Effect |
|--|--|
| E-step pow-dedup (drop dead `dpdf`) | **-35%**, bit-identical |
| `block_size` default 128->512 | **-18%** (compounds to ~-47%) |
| CUDA float64 (auto-selected) | **~4.5x** over 16-thread CPU (38.5 vs 172.8 ms/it), **numerically identical** (-3.42408) |

Full torch suite green (108 passed). Added `benchmarks/benchmark_gpu.py` (device/precision sweep, with warmup); findings in `.context/issue-63/perf_findings.md`.

### Honest caveats (the surprises)
- **float32 is NOT a free fast path.** It's 5-19x faster but **seed-flaky NaN** on full-size data (a mixture component underflows in float32's ~7-digit range ~iter 23; the degenerate-fit contract stops it loudly). Converges on small slices. Stabilization tracked in **#70**.
- **MPS is currently *slower* than CPU** on 32-channel data (0.51x) -- dispatch overhead unamortized on small montages; revisit with newer drivers / larger channel counts.
- **CPU threading is workload-limited** (~4 threads sweet spot on a laptop) -- documented as guidance, not forced (a library shouldn't mutate global `torch.set_num_threads`).
- Measurement note: a cold first CUDA call reads ~131 ms/it (2.5x inflation); always warm up before timing GPU.

Net: float64 CPU ~-47%, plus a clean 4.5x on GPU, all parity-preserving. The bigger multipliers (float32) need the stability work in #70.

neuromechanist:
## Review complete (3 Sonnet reviewers) + all findings addressed

**Code: clean.** All three reviewers independently verified the bit-identity -- one ran `torch.equal` on `_log_pdf_only` vs `_log_pdf_and_deriv`'s `log_pdf` (exactly equal, both branches) and confirmed the `|y|^rho` reuse is a true reuse (diffed vs main). No correctness bugs, no parity risk.

**Doc accuracy: several real fixes (my perf docs overstated).**
- `block_size=512` does **not** 'match the Fortran reference' -- Fortran's header default is 128 and it auto-tunes via `do_opt_block` (128-1024). Reworded.
- `benchmark_gpu.py` docstring called float32 'the production fast path (10-50x)', contradicting `perf_findings.md` (experimental, seed-flaky NaN). Aligned + fixed range to 5x CPU / 10-19x CUDA.
- Fixed the false 'all tests specify block_size explicitly' claim, the `test_block_size_independence`->`test_blocking_invariance` name, the overstated 'numerically identical' (->'agrees to 5 sig digits'), the unsupported cross-machine '~7x', and the untested workstation-threads claim.

**Tests added (test-analyzer, Priority 7):** `test_log_pdf_only_matches_log_pdf_and_deriv` (parametrized None/0/1/2/3/4) locks in the bit-identity for families 1-4 -- their only non-gated regression check -- plus `test_block_size_default_is_512`. 34 passed.

No findings skipped.

Issue #85, opened 2026-07-08
Phase 1: Native-Fortran adapter + CUDA host setup


neuromechanist:
Part of epic #84.

## Scope
Build native `amica` from the `.f90` source (gfortran + OpenMP) on a Linux x86 + CUDA host, and add a `_run_fortran` native-Fortran backend to the Phase B dimsweep harness (`benchmarks/benchmark_dimsweep.py`).

## Definition of Done
- [ ] Build script/docs for gfortran + OpenMP compile of `amica15.f90`/`funmod2.f90` on Linux x86.
- [ ] `_run_fortran` adapter: writes `.fdt` (Fortran-order) + param file, invokes the binary, parses the binary's own per-iteration timestamps for startup-immune per-iteration timing (not wall-clock of the whole process).
- [ ] Registered as a backend in `_BACKENDS` with availability detection (present only when a native build exists).
- [ ] Structurally validated on Mac (Rosetta `amica15mac`) and against the reverted Phase D prototype; real timing deferred to the Linux host.
- [ ] Real EEG only (channel subsets of ds002718). NO MOCKS.

neuromechanist:
Done in PR #88 (squash-merged into epic branch feature/issue-84-epic-xplat-benchmark). Native-fortran-f64 backend + cross-platform build validated on Mac arm64 and hallu x86 (RTX 4090).

Issue #87, opened 2026-07-08
Phase 3: Full matrix + CUDA multi-model cross-platform report


neuromechanist:
Part of epic #84.

## Scope
Run the whole grid on the CUDA host (torch-cpu / numpy / native-fortran / cuda) + Mac (mlx / mps); fill in the multi-model CUDA rows Phase B left pending; merge into a single cross-platform report + findings.

## Definition of Done
- [ ] Full grid executed on the Linux/CUDA host and Mac; results merged across platforms.
- [ ] Multi-model CUDA rows (left pending in Phase B / #77) filled in.
- [ ] Single cross-platform report + `.context/issue-84/` findings with honest caveats (native vs Rosetta, cross-machine GPU comparison, thread contention).
- [ ] Real EEG only. NO MOCKS.

neuromechanist:
Done in PR #93 (squash-merged into the epic branch). Cross-backend decomposition equivalence: all torch/MLX backends identical (f32==f64==CPU==CUDA==MPS==MLX), Fortran equivalent where well-determined; k-factor + EEGLAB-style variance-ordered de-sphered IC topomaps. Findings + figures in .context/issue-84/phase3_decomp_equivalence.md.

Issue #86, opened 2026-07-08
Phase 2: CPU core-count scaling sweep


neuromechanist:
Part of epic #84.

## Scope
Add a `--threads` sweep (4/6/8/12/...) to the dimsweep harness for the CPU backends (torch-cpu, numpy, native-fortran) with matched thread budgets, to see whether cores catch the GPU.

## Definition of Done
- [ ] `--threads` sweep axis wired to torch (`torch.set_num_threads`/interop), numpy (OMP/MKL env), and native-Fortran (OMP_NUM_THREADS) with matched budgets per run.
- [ ] Results captured per (backend x threads x dims) with per-iteration timing.
- [ ] Honest caveats on thread contention documented.
- [ ] Real EEG only. NO MOCKS.

neuromechanist:
Done in PR #89 (squash-merged into the epic branch). Complete cross-platform channel x core sweep, f64+f32, both machines; findings in .context/issue-84/phase2_cpu_scaling.md.

Issue #84, opened 2026-07-08
Epic: Cross-platform native-Fortran + CPU-scaling + CUDA benchmark


neuromechanist:
## Goal
A rigorous, cross-platform **result + performance** benchmark of every AMICA backend, extending
the Phase B harness (`benchmarks/benchmark_dimsweep.py`, #77) with the pieces intentionally left
out of epic #74: **native Fortran**, **CPU core-count scaling**, and **CUDA multi-model**. Answers
"how does the native reference and multi-core CPU compare to the GPU backends, and where are the
crossovers?"

## Why its own epic
Epic #74 (Apple-GPU acceleration) delivered the MLX backend and the single-machine Mac + CUDA
single-model comparison. This benchmark is a distinct, larger effort: it needs a native-x86 Linux
+ CUDA host (Apple's `amica15mac` is x86-under-Rosetta, and Mac CPU/Fortran timing proved
unreliable), a Fortran build from source, and multi-dimensional sweeps.

## Phases (proposed)
1. **Native-Fortran adapter + host setup.** Build amica from the `.f90` source (gfortran + OpenMP)
   on a Linux CUDA host; a `_run_fortran` adapter (write `.fdt` Fortran-order + param, parse the
   binary's own per-iteration timestamps for startup-immune timing). (A Mac Rosetta adapter was
   prototyped in the Phase D branch and reverted -- reuse it.)
2. **CPU core-count scaling.** `--threads` sweep (4/6/8/12/...) for the CPU backends
   (torch-CPU, numpy, native Fortran) with matched thread budgets, to see whether cores catch the GPU.
3. **Full matrix + CUDA multi-model.** Run the whole grid on the CUDA host (torch-cpu / numpy /
   native-fortran / cuda) + Mac (mlx / mps); fill in the multi-model CUDA rows Phase B left pending;
   merge into a single cross-platform report + findings.

## Constraints
Real EEG only (NO MOCKS; channel subsets of ds002718 for timing). Honest caveats (native vs
Rosetta, cross-machine GPU comparisons, thread-contention). Not run in CI (hardware-specific).

PR #88, opened 2026-07-08
Phase 1: Native-Fortran adapter + CUDA host setup

neuromechanist:
## Summary
Adds a `native-fortran-f64` backend to the Phase B dimension-sweep harness
(`benchmarks/benchmark_dimsweep.py`) that times an `amica15` compiled from source, plus a
cross-platform build (`benchmarks/fortran/`). First piece of epic #84: an honest
native-Fortran reference-timing row, on both the x86 CUDA host and Apple Silicon.

Closes #85
Part of epic #84

## What changed
- **`_run_fortran` adapter**: writes a real subset as a float32 column-major `.fdt`, renders a
  matched `input.param` (n_mix=3, pdftype=0, do_newton off, block_size 512, full fixed-iter
  budget), runs the binary, and times it from amica's **own per-iteration `out.txt` stamps**
  (startup-immune: after MPI init / data load / PCA / sphering). Registered in `_BACKENDS`;
  available only when a native binary exists (so CI / Apple-only checkouts are unaffected).
- **`benchmarks/fortran/build_amica.sh`** + **`vmath_shim.c`**: cross-platform native build.
  `amica15.f90` targets ifort + MKL, so a plain gfortran build needs three fixes, applied to a
  **build copy** (never the tracked reference source): `-cpp` (skip the MKL `#include`), a
  portable `random_seed` seed size, and a libm shim for AMD LibM's `vrda_exp`/`vrda_log`.
  Built clean on macOS arm64 (gfortran 16) and Ubuntu 24.04 x86_64 (gfortran 13).
- **Docs**: `benchmarks/fortran/README.md` (host setup, timing method + 10 ms resolution
  caveat, portability notes), dimsweep README/docstring updated, `.context/issue-84/`.

## Test plan
- `pyAMICA/tests/test_fortran_adapter.py` (real committed data, NO MOCKS): `.fdt` round-trip vs
  the canonical `load_data_file` (locks the column-major convention), channel-subset round-trip,
  `out.txt` parser (200 iters, LL), param rendering, and an end-to-end smoke that runs a native
  amica (skipped unless a binary is present / `AMICA_FORTRAN_BIN` is set).
- Validated end-to-end on **both** platforms: 5/5 pass against a native arm64 (Mac) and native
  x86 (hallu) binary. Real dimsweep on hallu (RTX 4090): native-fortran-f64 12.07 ms/it @32ch,
  44.83 @70ch vs torch-cuda-f64 34.88/38.61 and torch-cpu-f64 217/226; LL agrees ~2 digits.
- `ruff check` + `ruff format` clean; `shellcheck` clean; full suite collects (154 tests).

## Notes / caveats
- Per-iteration stamp is ~10 ms resolution; the adapter rejects configs that round to 0.00 s/iter
  and documents the floor (use per-iter >> 10 ms, i.e. the 70-ch size).
- native-Fortran uses libm (not vendor SIMD) math -> portable-baseline timing, not max-tuned.
- Build portability fixes are candidates to upstream to sccn/amica (#44 M1 compile, #49 Ubuntu).
- Out of scope: `--threads` CPU-scaling sweep (Phase 2 #86); full grid + CUDA multi-model report
  (Phase 3 #87).

### DIFF
--- a/.context/issue-84/phase1_build_notes.md
+++ b/.context/issue-84/phase1_build_notes.md
@@ -19,8 +19,9 @@ reference is never modified):
 
 Also `-std=legacy -fallow-argument-mismatch -ffree-line-length-none`. Built clean on both
 **macOS arm64 (gfortran 16, brew open-mpi + lapack)** and **Ubuntu 24.04 x86_64
-(gfortran 13, apt openmpi + liblapack-dev)**. Upstream target: sccn/amica #44 (M1 compile),
-#49 (Ubuntu recompile).
+(gfortran 13, apt openmpi + liblapack-dev)**. Upstream target: sccn/amica #44 (compile on
+M1) is an exact match; #49 asks about Ubuntu builds too, though via Intel OneAPI rather than
+this gfortran path.
 
 ## Timing method
 
--- a/.context/issue-84/phase1_build_notes.md
+++ b/.context/issue-84/phase1_build_notes.md
@@ -1,0 +1,49 @@
+# Phase 1 (#85): native-Fortran adapter + build notes
+
+Adds a `native-fortran-f64` backend to `benchmarks/benchmark_dimsweep.py` that times an
+`amica15` compiled from source, plus a cross-platform build (`benchmarks/fortran/`).
+
+## gfortran build portability (source targets ifort + MKL)
+
+`amica15.f90` assumes Intel's toolchain. A plain `gfortran` + LAPACK build needs three
+fixes, applied by `build_amica.sh` to a **build copy** (the tracked `pyAMICA/amica15.f90`
+reference is never modified):
+
+1. `-cpp` -- resolve the `#ifdef MKL` guards so `include 'mkl_vml.f90'` is skipped.
+2. `random_seed` -- the source's size-2 seed array (ifort-sized) is rejected by gfortran
+   (wants 8); the build copy uses a portable default seed. Timing-neutral (seeding was
+   already clock-based / non-reproducible).
+3. `vmath_shim.c` -- the non-MKL branch calls AMD LibM `vrda_exp`/`vrda_log`; the shim
+   provides them as libm loops so no vendor math lib is needed. IEEE-accurate; a vendor
+   SIMD math lib could make the exp/log-heavy E-step somewhat faster.
+
+Also `-std=legacy -fallow-argument-mismatch -ffree-line-length-none`. Built clean on both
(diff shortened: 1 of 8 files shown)


neuromechanist:
## Review summary (4 agents, Sonnet)

Ran code, tests, silent-failure, and comment reviews. **Code correctness: clean** (byte order, shim ABI, regex, param ordering all verified against real data/hardware). Addressed all actionable findings in b2df278:

**Silent failures**
- [x] Native crash now prints full stderr/stdout before raising (`main()` truncates the exception to ~70 chars, so the transcript is the only place to see LAPACK/MPI/segfault causes).
- [x] Explicitly-requested-but-unavailable backends now emit a `WARNING` instead of silently dropping (was: sweep exits 0 having benchmarked fewer backends than asked).
- [x] `build_amica.sh` verifies the `random_seed` sed-patch actually applied (`grep -q`, else fail loud) so source drift can't produce a binary that only crashes at runtime.
- [x] `_parse_fortran_out` uses `errors="replace"` (corruption surfaces as U+FFFD, breaks the regex cleanly) instead of `errors="ignore"`.
- Skipped (with rationale): the regex not matching a negative seconds value is theoretical (monotonic timer); leaving it excluded is the safe behavior.

**Test coverage** (all error-guard paths were previously untested)
- [x] Added: nonzero-exit raise (real system `false`, unconditional), sub-10ms zero-guard raise (verified repro 8ch x 2000), `<2` timed-iters raise, `_fortran_available`/`_available` dispatch + share-gate, and multi-model param rendering. `pcakeep 16` now asserted. 4 -> 10 tests; binary-gated ones skip on CI/Apple-only.
- [x] Fixed the flaky "16ch/8k rounds to 0.00s" docstring claim (empirically alternates 0.00/0.01; the real reliable sub-floor size is 8ch x 2000).

**Comment/doc accuracy**
- [x] Fixed two factual bugs: the "not on Apple/CI" docstring (the Apple native build works) and the "amica15 is single-model" comment (it supports `num_models>1`; the adapter only gates `share`). Aligned `.gitignore` + `README_dimsweep.md`.
- [x] Documented the native-fortran `final_ll` reproducibility caveat (clock-seeded, unlike the fixed-seed backends) in `README_dimsweep.md`.
- [x] Softened the uncited "equal or better than vendor" accuracy claim in `vmath_shim.c`; qualified the sccn/amica #49 reference (OneAPI, not gfortran).

Full suite: 10 passed (both with and without a native binary); ruff + shellcheck clean.

PR #100, opened 2026-07-11
EEGLAB drop-in output parity (#92)

### LINKED ISSUE #92: Drop-in EEGLAB AMICA replacement: equivalent output, no extra interpretation
## Goal
pyAMICA (and the native-Fortran build path) should be a **drop-in replacement for EEGLAB's AMICA**: the output must be directly equivalent, requiring no additional interpretation or re-ordering by the user. Phase 3 (#87) proved the *decomposition* is equivalent (backends recover the same ICs where well-determined); this issue tracks the *output/convention* parity that makes it a true replacement.

## Requirements
- **Variance-ordered ICs** by back-projected variance (EEGLAB convention, IC1 = highest variance), so component indices mean the same thing as in EEGLAB. (Started in #87's figure ordering.)
- **Output format parity**: write results EEGLAB's `loadmodout15` / `pop_runamica` can read (W, A, S/sphere, mean, c, LL, comp order) so a pyAMICA run drops into an existing EEGLAB workflow.
- **Sign/scale conventions** matching EEGLAB so topomaps and activations line up without manual flips.
- **Same reduced-montage behavior** (see #91: spatially-distributed channel subsets).
- A documented round-trip: run pyAMICA -> load in EEGLAB -> identical component maps/order.

## Why
The value proposition is "use pyAMICA/GPU instead of EEGLAB AMICA and get the same answer, faster." Any step that needs manual reinterpretation (re-sorting ICs, flipping signs, reformatting output) undercuts that. Related: #87 (equivalence), #90 (data-size sweep), #91 (distributed channels).

neuromechanist:
Closes #92.

## Summary
Makes pyAMICA a drop-in replacement for EEGLAB's AMICA: a PyTorch NG fit can now
be written as the exact `loadmodout15` binary output directory, so it loads in
EEGLAB with the components in the same order and orientation, with no manual
re-sorting, sign-flipping, or reformatting.

Previously only the legacy NumPy backend could write this format; the primary
PyTorch backend could only `torch.save` a private dict.

## What changed
- **Shared writer** `write_amicaout()` in `numpy_impl/load.py` (co-located with its
  `loadmodout` reader counterpart): emits `gm, W, S, mean, c, alpha, mu, sbeta,
  rho, comp_list, LL`. `numpy_impl/core.py:_write_results` now delegates to it
  (byte-identical, no duplication).
- **`AMICATorchNG.write_amica_output(outdir)`** and the **`AMICA.write_amica_output`**
  wrapper (with the `_check_usable` degenerate-fit guard, #50).
- **`variance_order()`** on backend + wrapper: EEGLAB back-projected-variance order
  (IC1 = highest), using the same `W` convention real MATLAB `loadmodout15.m` uses.
- **Docs:** `docs/guides/eeglab.md` (EEGLAB interop + round-trip), wired into the nav.

## Validation (real data + Fortran binary, no mocks)
- NumPy-vs-Fortran parity tests still pass (writer refactor is byte-identical).
- New round-trip tests (`test_amica_ng_wrapper.py`): byte-fidelity of every written
  file, `loadmodout` readability + shapes, variance ordering, and the requires-fit
  guard. Full non-slow suite green (166 passed).
- **MATLAB `loadmodout15` round-trip (column-major, exactly as MATLAB reads):** on
  a converged fit (LL -3.399), `variance_order()` equals the reader's `origord`,
  svar is descending, and per-component mixture proportions sum to 1.0 -- i.e. the
  written directory is read correctly by real EEGLAB with the components in the
  right order. Matched-config IC equivalence vs Fortran stays the ~0.997 from
  `validate_implementations.py`.
- **A format defect was caught and fixed here** (see review-response comment): the
  mixture params were written C-order but EEGLAB/Fortran store column-major, so
  MATLAB would have read scrambled mixture proportions. Fixed writer + both numpy
  readers to Fortran order; single-model output is now byte-faithful to Fortran.

## Notes
- Single-model output is byte-identical to the Fortran reference. Multi-model output
  is self-consistent (round-trips through `loadmodout15`) but not byte-identical to a
  native multi-model AMICA run (issue #27 scope).
- The Python port `loadmodout` reads C-order where MATLAB reads column-major (a
  pre-existing latent transpose in the reader, out of this writer's scope); recorded
  in `.context/scratch_history.md`. The written bytes are correct for real EEGLAB.

Tested: yes (see Validation).

### DIFF
--- a/.context/scratch_history.md
+++ b/.context/scratch_history.md
@@ -45,6 +45,27 @@ python -m pyAMICA.amica_cli pyAMICA/sample_data/sample_params.json --verbose --o
 gfortran -O3 -fopenmp amica17.f90 funmod2.f90 -o amica -llapack -lblas
 ```
 
+## Issue #92 (EEGLAB drop-in output): loadmodout reader quirks (pre-existing, out of writer scope)
+The writer (`write_amicaout` in `numpy_impl/load.py`, used by both backends via
+`torch_impl/core.py:write_amica_output` and `numpy_impl/core.py:_write_results`)
+emits bytes byte-identical to Fortran for a single model, so the REAL MATLAB
+`loadmodout15.m` reads it correctly (it reads column-major -> recovers the true
+unmixing `W_true = self.W.T`, so its `A = pinv(W_true @ S)` and its variance
+ordering match the model's native components; `AMICATorchNG.variance_order` uses
+the same `W_fort = self.W.T` and so matches real EEGLAB).
+- The Python PORT `numpy_impl/load.py:loadmodout` reads the `W` file **C-order**
+  (`W.reshape(nw,nw,m)`), where MATLAB reads column-major. So the port's `mod.W`
+  is the transpose of MATLAB's, and its `A`/`svar`/`origord` use `pinv(self.W_int
+  @ S)` (untransposed) -- a latent transpose vs `loadmodout15.m`. Masked because
+  every parity test matches `W` via transpose-tolerant Hungarian |corr|, and
+  nothing consumes `.A`/`.svar`/`.origord`. Do NOT "fix" by transposing the reader
+  without re-checking #24/#37 locked parity and the numpy write/read round-trip.
+- The port also stores `svar` in fit order (not sorted); `loadmodout15.m` stores
+  it descending. `svar[origord]` is descending. Tests assert on `svar[origord]`.
+- Consequence: `variance_order()` (native/EEGLAB-correct) does NOT equal the
+  port's `origord` (transposed convention). Validate `variance_order` against the
+  real MATLAB `loadmodout15` round-trip, not against the numpy port's `origord`.
+
 ## Lessons / check first next time
 - [ ] Positive LL almost always means a wrong PDF normalization constant.
 - [ ] NaN at a phase transition (e.g. Newton start) points at unclipped gradients or zero denominators.
--- a/.context/scratch_history.md
+++ b/.context/scratch_history.md
@@ -45,26 +45,29 @@ python -m pyAMICA.amica_cli pyAMICA/sample_data/sample_params.json --verbose --o
 gfortran -O3 -fopenmp amica17.f90 funmod2.f90 -o amica -llapack -lblas
 ```
 
-## Issue #92 (EEGLAB drop-in output): loadmodout reader quirks (pre-existing, out of writer scope)
-The writer (`write_amicaout` in `numpy_impl/load.py`, used by both backends via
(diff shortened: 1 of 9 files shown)


neuromechanist:
## Review round 1 (Sonnet) + a MATLAB round-trip surfaced a real format bug

The `/manuscript`-grade "no shortcuts" standard paid off here: verifying against real MATLAB `loadmodout15` exposed a genuine format defect that the Python-only tests missed.

### Format bug (found + fixed)
pyAMICA wrote the non-square mixture params (`alpha`/`mu`/`sbeta`/`rho`, shape `(num_mix, num_comps)`) and `c`/`comp_list` in **C-order**, but Fortran/EEGLAB store them **column-major**. So real MATLAB `loadmodout15.m` (column-major reads) would have read *scrambled* mixture params, breaking the variance ordering and density model. Diagnostic: reading the committed genuine-Fortran `amicaout/alpha` column-major gives per-component mixture proportions summing to exactly 1.0; C-order gives garbage `[0.49, 1.31]`.

Fixed the writer (`write_amicaout`) and both numpy readers (`loadmodout`, `data.py:load_results`) to use Fortran (`order="F"`) layout for those arrays. `W` (square) stays C-order and is byte-identical to Fortran (the #24 transpose cancels against column-major storage); `S` is symmetric; `mean`/`gm`/`LL` are 1-D. This also fixes a latent bug where the numpy reader mis-read genuine Fortran mixture params.

**Verification:** re-ran the live MATLAB-faithful (column-major) round-trip on a converged fit (LL -3.399): `variance_order()` now equals the reader's `origord`, svar is descending, and mixture proportions sum to 1.0. Matched-config IC equivalence remains the ~0.997 from `validate_implementations.py`.

### Review findings (all addressed)
1. **Important - LL/keep_best mismatch:** when the #51 safeguard restores an earlier iterate, the written `LL` file now ends at that kept iterate (`LL[-1] == final_ll_`), not at a later discarded overshoot. New deterministic test `test_write_amica_output_ll_matches_kept_iterate` (n_models=2 run that reliably triggers a restore).
2. **Important - guard-test gap:** `write_amica_output` and `variance_order` added to `test_degenerate_fit_refuses_output`; new `test_variance_order_requires_fit`.
3. **Suggestion - tautological ordering assertion:** replaced with a meaningful column-major check (mixture proportions per component sum to 1 after the loadmodout round-trip).

### Regression safety
NumPy-vs-Fortran parity suite green (the reader change affects how genuine Fortran `amicaout` is read); full torch wrapper suite green (16 passed); byte-fidelity test updated to read the column-major arrays with `order="F"`.

Issue #77, opened 2026-07-08
Phase B: Dimension-sweep benchmark - Apple GPU vs CPU crossover


neuromechanist:
## Context
Measurement phase of the epic. MPS lost on 32-channel data (#63: 0.51x) because of
**fixed per-op Metal dispatch overhead**, which amortizes as tensors grow - 32 channels
was the wrong regime to judge MPS. This phase finds the crossover where Apple GPU actually
wins. Runs last so it can benchmark the Phase A float32 path (and Phase C MLX if landed).

## Approach
- Extend `benchmarks/benchmark_gpu.py` with a **dimension sweep**: channel count
  (32/64/128/256), `block_size`, and `n_models`.
- Compare MPS-float32 (and MLX if Phase C landed) vs CPU float64/float32 - warmed, min of
  repeats, CUDA-synchronized.
- Identify and document the crossover point(s).

## Data note (timing, not correctness)
Real sample EEG is 32 channels; higher channel counts are produced by tiling/replicating
real channels purely to size the tensors. This is a **throughput benchmark, not a
correctness test** - the NO-MOCKS rule governs correctness tests, and no correctness claim
is made from tiled data.

## Acceptance
- Crossover curve documented in `.context/issue-63/perf_findings.md` (or a new perf note).
- `benchmark_gpu.py` runs the sweep behind a single CLI flag.

## Dependencies
Blocked by **Phase A** (needs stable float32); benefits from **Phase C** (compare MLX).
Research: `.context/mps_pathways.md` Pathway B.

neuromechanist:
Done via PR #80 (squash-merged into the epic branch, epic #74 Phase B). Cross-platform benchmark on real 70-ch EEG: MLX is the Apple-GPU win (~15-25 ms/it, ~7x over torch-CPU, faster than an RTX 4090 at EEG scale); PyTorch-MPS never wins (>= CPU); results agree across cpu/mps/cuda/mlx to ~3 digits. Multi-model has no GPU path yet -> follow-up #81. See .context/issue-77/benchmark_findings.md.

PR #112, opened 2026-07-11
Parity & performance tables in paper + full results in docs

### LINKED ISSUE #111: paper+docs: parity & performance tables
The JOSS paper claims Fortran parity in prose but shows no numbers and does not make explicit where parity holds (single-model, bit-exact) versus where it is fundamentally limited (multi-model AMICA is not partition-identifiable). Add:

- **paper.md**: two compact summary tables (parity/accuracy + per-backend throughput) and a pointer to the full benchmark tables and figures in the docs.
- **docs/guides/validation.md**: the full measured tables (single- and multi-model ms/iteration across channel counts, cross-backend log-likelihood agreement, the multi-model ensemble-equivalence distribution + plot, and the k=152 precision split) so the paper stays concise while the complete results live in the documentation.

All numbers are the repo's real measured benchmark results (issues #77, #27, #90); no synthetic data.

neuromechanist:
Closes #111.

Makes the parity claim concrete (and honest about where it holds vs. where it is fundamentally limited), and banks on the float32 unlock.

## paper.md
- Two compact tables in *Implementation and validation*: **parity** (single-model LL/correlation/bit-exactness + the multi-model distributional-equivalence result) and **throughput** (per-backend ms/iteration on real 70-ch EEG). Trimmed the now-duplicated inline numbers.
- Summary now banks on **float32**: pyAMICA adds single-precision execution the CPU-only reference lacks (5-19x faster, required on Apple GPUs), with an honest caveat that float32 is 7-significant-digit, not bit-exact, so float64 stays the parity default.
- Points readers to the full tables/figures in the docs.

## docs/guides/validation.md
- Full measured tables: single- and multi-model ms/iteration across 16-70 channels, cross-backend log-likelihood agreement, the k=152 precision split (native-Fortran f64 vs CUDA f64 = 0.995), and the multi-model ensemble-equivalence distribution (within-Fortran / within-pyAMICA / between + Mann-Whitney/TOST) with the ensemble plot.
- Enabled `pymdownx.blocks.caption`, which fixes a pre-existing bug where `/// caption` blocks (the k-factor figure) rendered as literal text; both figures now have real captions.

## Data provenance
Every number is the repo's real measured benchmark result (issues #77 dimsweep, #27 multi-model ensemble, #90 k-sweep) on real EEG (sample data / OpenNeuro ds002718). No synthetic data.

Verified: `mkdocs build --strict` passes (exit 0, no warnings); the JOSS paper PDF builds via the draft-pdf workflow (checked on this PR).


neuromechanist:
## Peer-review panel (4 independent Sonnet reviewers) — disposition

Ran a general reviewer plus a 3-lens panel (statistics, methods/reproducibility, novelty/significance) via the manuscript:paper-review skill. All four converged on two blockers and a set of structural fixes, all addressed in commit `dd26ff3`:

**Critical — fixed**
- **float32 overclaim.** "5-19x faster / float32 unlocks the GPU speedups" was contradicted by the paper's own Table 2 (CUDA float32 ~= float64; MPS float32 slower than CPU), confirmed by the rigorous cross-platform benchmark (issue-84: GPU is overhead-bound, "f32 is no faster"). Reframed to the accurate story: float32 is numerically faithful (4-5 sig digits vs float64), required on Apple GPUs (enabling the fastest MLX backend), and CUDA is overhead-bound so float32 adds little there.
- **Multi-model pseudoreplicated p-values.** The Mann-Whitney/TOST were computed over pairwise correlations (190/400) from only 20 independent runs, so the p-values are not valid inference. Removed them from the paper; the multi-model equivalence now rests on the valid descriptive overlap (three distribution means within 0.01, sd ~0.05, Figure 1). The docs keep the tests with an explicit pairwise-dependence caveat.

**Major — fixed**
- Scoped "bit-level agreement" to the score-function algebra; "byte-identical" -> "byte-identical in layout" (file format, not solution).
- Rescoped "parity" to single-model; Table 1 recaptioned "Single-model parity and multi-model distributional equivalence".
- Added the previously-omitted multi-model log-likelihood caveat (differs slightly at matched budget; converges to Fortran's mean with ~2x iterations).
- Reconciled 0.997 (sample EEG, single-model) vs 0.98/0.995 (cross-backend sweep, larger recording) by stating each config.
- Table 2: labeled single-model, added settings + warm/min-of-repeats + reproducibility caveat (external data + specific hardware).
- Embedded the multi-model ensemble figure (Figure 1) in the paper.
- Added repository + Zenodo DOI availability line; added a COI/author-contribution note (S.M. co-developed AMICA; A.D. leads EEGLAB).
- State of the field: reworded from sole-possession to "more rigorous parity validation than either alternative publishes" (herforth ships a Fortran-comparison test); clarified pyAMICA offers sklearn API + EEGLAB I/O, not an MNE-Python wrapper.

**Minor — fixed**: hedged/attributed the dipolarity claim; "three sig digits" -> "max pairwise difference ~0.003"; glossed Hungarian matching; added access dates to the two @misc citations.

**Verification**: JOSS paper PDF builds (draft-pdf workflow); `mkdocs build --strict` passes. All table numbers trace to real benchmark records (issues #77/#27/#90/#84).

**Noted, not blocking**: Figure 1's embedded panel annotations still show the old p-values (the PNG predates this revision; regenerating needs the ensemble data). The caption and text frame it descriptively; regenerating the figure is a future polish. Funding acknowledgement left for the author to add if applicable.

Issue #117, opened 2026-07-13
validate_implementations.py: fix type-safety diagnostics


neuromechanist:
Static analysis (Pyright) flags real issues in `validate_implementations.py`:

1. `run_fortran_amica` is typed `-> Dict` but returns `None` on several early-exit
   paths (binary missing, non-zero return code, timeout, exception). Callers
   already handle this (`compare_results(fortran_results: Optional[Dict], ...)`),
   so the annotation is just wrong. Fix: `-> Optional[Dict]`.
2. Inside `run_fortran_amica`, `import os` and `original_dir = os.getcwd()` happen
   inside the `try:` block; the `except` branches reference both, so Pyright
   correctly flags them as possibly-unbound if the import itself ever failed.
   Fix: hoist `import os` to the top-level imports and set `original_dir` before
   the `try:`.
3. `load_eeglab_data(..., dtype: np.dtype = np.float32)` in
   `pyAMICA/torch_impl/utils.py` has a default value that does not match its own
   annotation (`np.float32` is a scalar type, not an `np.dtype` instance). Fix:
   widen the annotation (e.g. `npt.DTypeLike`).

Verified NOT real bugs (false positives, no fix needed): the `Import "torch"` /
`Import "scipy.optimize" could not be resolved` diagnostics are a Pyright/IDE
environment issue -- both import and run fine under `uv run python -c "import
torch, scipy"`. Leaving those as-is.

PR #120, opened 2026-07-13
Add Amari distance metric and Fortran/Apple-CPU perf rows

### LINKED ISSUE #116: paper+docs: Fortran/Apple-CPU rows in Table 2 + Amari-distance detail
Follow-up to #111/#112.

## Table 2 (paper.md) + docs performance tables
Table 2 has no Fortran row. We already have the numbers from the #86 CPU
core-scaling sweep (`.context/issue-84/phase2_cpu_scaling.md`):
- Native Fortran, Intel Xeon/RTX4090 workstation, 24 cores: 30.0 ms/iter (70ch)
- Native Fortran, Apple Silicon, 8 cores: 70.0 ms/iter (70ch)
- Apple Silicon torch-cpu f64/f32 at the same 8-core point (for a fair Fortran vs
  torch comparison on the same machine), also in that file.

Add these to paper.md Table 2 and reproduce the full core-scaling tables (both
platforms) in docs/guides/validation.md, since the paper explicitly points there
for full results.

## Amari distance detail for multi-model equivalence
Table 1's multi-model row only reports aggregate mean/sd of Hungarian-matched
correlation (from the #27 N=20-each ensemble). Add the Amari distance (Amari,
Cichocki & Yang 1996), a standard ICA unmixing-matrix comparison metric that is
permutation/scale-invariant by construction (complementary check to the
Hungarian-matched correlation, no Hungarian assignment step needed).

No re-fitting required: `.context/issue-27/ensemble.npz` already has the 20
Fortran + 20 NG stacked unmixing matrices (real EEG fits, not synthetic) needed
to recompute this metric directly.

Add:
- `amari_distance()` implementation + within-Fortran/within-NG/between summary
  stats (paper.md Table 1 row + docs).
- A per-run detail table (not just the aggregate) in docs/guides/validation.md,
  giving each run's mean within-group and between-group value for both metrics
  (correlation and Amari distance).
- A test on the real saved ensemble data (metric properties: self-distance 0,
  permutation/scale invariance; no mocks).

neuromechanist:
Closes #116.

## Amari distance
`amari_distance()` in `validate_implementations.py` is a permutation- and
scale-invariant unmixing-matrix comparison metric (Amari, Cichocki & Yang,
1996), used alongside (not instead of) the existing Hungarian-matched
correlation as a second, assignment-free conformity check. Recomputed from
the already-saved `.context/issue-27/ensemble.npz` matrices (real EEG fits
from the #27 study, no re-fitting needed):

- Single-model: ~0.008 (matches the ~0.997 correlation).
- Multi-model: within-Fortran 0.174, within-pyAMICA 0.155, between 0.161
  (sd ~0.02), run-level permutation p=0.998 -- agrees with the
  correlation-based "between not worse than within-Fortran" conclusion.

Per-run detail (each of the 40 runs' own within/between mean for both
metrics) is a new table in `docs/guides/validation.md`, plus
`.context/issue-27/per_run_detail.csv` (regenerable via
`.context/issue-27/amari_distance.py`).

## Table 2 / performance tables
Table 2 (paper.md) had no Fortran row. Added native Fortran rows (Intel
24-core workstation, Apple Silicon 8-core) from the existing #86 CPU
core-scaling sweep, plus the full core-scaling comparison table (both
platforms, matching Fortran against torch-CPU/NumPy at the same core counts)
in the docs.

## Formatting
Both `paper.md` and `docs/guides/validation.md` are reflowed to semantic line
breaks per the documented convention. Content is unchanged -- verified via
whitespace-collapsed diff before and after, plus a pandoc PDF build and an
`mkdocs build --strict` for the docs page.

## Test plan
- [x] `uv run pytest pyAMICA/tests/test_amari_distance.py` (new: self-distance
  zero, permutation/scale invariance, nonzero for unrelated matrices) passes
- [x] `uv run ruff check` / `ruff format --check` pass
- [x] `uv run ty check` passes on all touched Python files
- [x] `pandoc --citeproc paper.md` builds a PDF with all citations resolving
- [x] `mkdocs build --strict` builds the docs site with no warnings


neuromechanist:
## Review findings and fixes

Ran code-review + test-coverage-analysis on this PR. Summary:

**Critical (fixed):** `amari_distance(Wa, Wb)` was not symmetric under argument
swap (verified: ~0.03 difference for swapped arguments on random matrices).
`.context/issue-27/amari_distance.py` called it in three different argument
orders across `pairwise()`, `per_run_detail()`, and `perm_test_not_worse()`,
so the JSON summary and the per-run CSV didn't reconcile (the "between" mean
differed by ~0.004 depending on which table you read). Fixed by making
`amari_distance` average both directions (a proper symmetric distance);
recomputed all downstream numbers (paper.md, docs, JSON, CSV) with the fix.
New values: within-Fortran 0.174, within-pyAMICA 0.154, between 0.163
(was 0.161), permutation p>0.999 (was 0.998); single-model 0.006 (was 0.008).
Verified the CSV and JSON now reconcile exactly.

**Important (fixed):** `amari_distance` silently produced NaN for a
degenerate all-zero row/column or an n<2 matrix (no guard, no exception).
Added explicit `ValueError`s for both cases, plus tests.

**Important (documented):** `model_amari`'s try-both-pairings model-label
resolution has a non-uniform bias magnitude across groups (~0.02-0.03,
comparable to the within-Fortran/within-pyAMICA gap). Added a caveat sentence
in docs/guides/validation.md noting this contributes to some of the observed
spread, rather than treating the gap as purely a convergence-speed effect.

**Test coverage (fixed):** added `pyAMICA/tests/test_amari_ensemble_script.py`
covering `model_amari`'s pairing selection and `perm_test_not_worse`'s
`higher_is_worse` sign-flip logic on synthetic data (previously untested;
a sign bug there could have silently inverted the reported conclusion).

**Not changed (false positive / out of scope):** duplication between
`.context/issue-27/amari_distance.py`'s local `xcorr`/`pairwise` and
`multimodel_ensemble.py`'s versions -- low-risk, point-in-time analysis
scripts, consistent with existing repo convention.

All tests pass (`uv run pytest pyAMICA/tests/test_amari_distance.py
pyAMICA/tests/test_amari_ensemble_script.py`), ruff/ty clean, pandoc PDF and
`mkdocs build --strict` both verified with the corrected numbers.

neuromechanist:
## Follow-up: single-model Amari distance now has a real test

A second review pass caught that the single-model "Amari distance ~0.008"
claim (now ~0.006, since the symmetry fix above) had no backing computation
anywhere in the repo -- unlike every other number in this PR, which traces to
a committed, regenerable artifact.

Fixed: added an \`amari_distance\` assertion (\`< 0.05\`, actual ~0.006) to the
existing \`test_end_to_end_correlation_vs_fortran\` in
\`pyAMICA/tests/torch_tests/test_ng_backend.py\`, right alongside the existing
correlation assertion that already backs the ~0.997 figure. This also
surfaced (via \`ty\`, now that \`run_fortran_amica\` is correctly typed
\`Optional[Dict]\` per #117) a missing None-check before subscripting the
Fortran result -- added an explicit \`assert fortran is not None\`.

Verified: \`uv run pytest pyAMICA/tests/torch_tests/test_ng_backend.py::test_end_to_end_correlation_vs_fortran -m slow\` passes locally (macOS, real Fortran binary).

neuromechanist:
## Skipped (intentional, not a false positive)

Both reviewers flagged that \`.context/issue-27/amari_distance.py\`'s local
\`xcorr\`/\`pairwise\` duplicate the same functions in the pre-existing
\`.context/issue-27/multimodel_ensemble.py\`, and that the \`importlib\`-based
loader pattern is repeated across the new script and its test. Skipping:
both are one-off, point-in-time analysis scripts under \`.context/\`, not
package code (no import path exists between them without adding one), this
matches the repo's existing convention for this class of script (no shared
module for \`.context/\` analysis helpers today), and the low-risk assessment
from both reviews (confidence ~40-50) agrees this isn't worth the churn of
introducing a new shared module for two files.

Issue #123, opened 2026-07-13
numpy_impl: do_reject outlier rejection is non-functional (data_mask never initialized/consumed)


neuromechanist:
Found while fixing ty diagnostics (#122). \`AMICA_NumPy._reject_outliers()\`
(\`pyAMICA/numpy_impl/core.py\`) references \`self.data_mask\` at
\`self.data_mask[outliers] = False\`, but \`data_mask\` is never initialized
anywhere in \`__init__\`/\`_initialize_parameters\` -- \`ty\` correctly flags this
as \`Object of type Self has no attribute data_mask\`. If a user sets
\`do_reject=True\` with \`maxrej > 0\` and the iteration reaches the reject
schedule (\`_optimize\`'s do_reject block), this raises \`AttributeError\` at
runtime.

Separately, even once \`data_mask\` exists, nothing reads it:
\`_get_updates_and_likelihood\`/\`_get_block_updates\` iterate over
\`self.data\` directly with no reference to \`data_mask\` at all, unlike the
torch backend's \`good_idx\` mechanism (\`AMICATorchNG\`, which actually
restricts \`X_use = X_t[:, self.good_idx]\` each iteration). So the rejection
mask, even if computed without crashing, currently has zero effect on the
fit -- the feature is a no-op at best, a crash at worst.

No test exercises \`do_reject\` on the NumPy backend (only the torch backend's
\`test_ng_backend.py\`/\`test_ng_pdf_families.py\` do), which is presumably why
this has gone unnoticed.

## Scope
Port the torch backend's \`good_idx\`-style filtering to the NumPy backend:
initialize a good-sample index/mask in \`_initialize_parameters\`, restrict the
per-iteration data to it in \`_get_updates_and_likelihood\`, and shrink it in
\`_reject_outliers\` (mirroring \`AMICATorchNG._reject_outliers\`). Add a
NumPy-backend test analogous to the torch \`do_reject\` tests.

Not fixed here: #122 only added a minimal \`data_mask\` initialization to stop
the immediate crash risk; it does not make the feature functional.

neuromechanist:
Correction to the root-cause scope (surfaced by the PR #124 type-safety review):

[...]

Issue #119, opened 2026-07-13
CI: enforce ty type-checking + add pre-commit config (ruff + ty)


neuromechanist:
Currently `.github/workflows/ci.yml` runs `ruff check` + `ruff format --check`
as a hard gate (the `lint` job), but never runs `ty` at all -- type-checking is
untested in CI. There is also no `.pre-commit-config.yaml` in this repo, so
neither tool runs locally pre-commit (per the global dev-standards
`pre-commit` template).

## Scope
1. Add a `ty check .` step to the `lint` job in `.github/workflows/ci.yml`
   (or a new `typecheck` job alongside it).
2. Add `.pre-commit-config.yaml` with `ruff check --fix`, `ruff format`, and
   `ty check` hooks, scoped to staged files only (per project convention:
   "pre-commit hooks should check staged files only").

## Blocker to resolve first
`uv run ty check .` currently reports **~496 pre-existing diagnostics**
across the repo (mostly in `pyAMICA/torch_impl/core.py`, e.g. `self.W`/
`self.sphere` typed as `None | Tensor` and subscripted/used without narrowing).
Turning `ty check` into an immediate hard CI gate would fail the build on
`main` as-is. Options, pick one before wiring the gate:
- (a) Clean up the existing diagnostics first (likely its own separate,
  larger issue -- `torch_impl/core.py` is the bulk of them), then gate.
- (b) Land `ty check` as informational/non-blocking in CI now (so regressions
  are visible in logs without breaking builds), track cleanup separately,
  flip to blocking once clean.

Not scoping the 496-diagnostic cleanup into this issue; decide (a) vs (b) and
size that work separately.

PR #124, opened 2026-07-14
Fix ty type-check diagnostics repo-wide (496 -> 0)

### LINKED ISSUE #122: Fix ty type-check diagnostics repo-wide (blocks #119)
Prerequisite for #119 (enforcing \`ty check\` in CI + pre-commit): \`uv run ty check .\`
currently reports 496 diagnostics. Root cause for the large majority: stateful
estimator classes (\`AMICATorchNG\`, \`AMICA_NumPy\`, MLX backends) initialize
fit-dependent attributes (\`self.W\`, \`self.A\`, \`self.sphere\`, \`self.c\`,
\`self.comp_list\`, ...) to \`None\` in \`__init__\` and only assign real tensors/
arrays in \`fit()\`. ty correctly infers these as \`Tensor | None\` (etc.) and
flags every downstream use (inside the class's own methods, and in test files
that access \`model.W\`/\`model.A\` directly after \`fit()\`) as a potential
None-subscript/None-attribute/wrong-argument-type error.

Breakdown (excluding \`.venv\`/stdlib stub false positives):
- \`pyAMICA/torch_impl/core.py\`: 170
- \`pyAMICA/tests/torch_tests/test_ng_sharing.py\`: 107
- \`pyAMICA/numpy_impl/core.py\`: 91
- \`pyAMICA/tests/torch_tests/test_ng_backend.py\`: 87
- \`pyAMICA/tests/torch_tests/test_ng_pdf_families.py\`: 35
- \`pyAMICA/numpy_impl/data.py\`: 27
- \`pyAMICA/mlx_impl/core.py\`: 24
- remainder scattered across smaller files/tests

## Fix approach
Not renaming the public \`W\`/\`A\`/\`sphere\`/... attributes (breaking API change,
used directly by tests/docs) and not silencing with \`cast()\`/\`# type: ignore\`
(defeats the point, matches none of this project's "no silent failures"
ethos). Instead: narrow with \`assert x is not None\` at point of use --
internal EM-loop call sites treat this as an enforced invariant ("trust
internal code", per CLAUDE.md); public accessor methods
(\`get_unmixing_matrix\`, \`get_mixing_matrix\`, \`transform\`, \`save\`, ...) get a
real \`RuntimeError\` guard since a user could call them before \`fit()\`; test
files get an \`assert\` right after the relevant \`fit()\` call, which also
documents "fit succeeded" as a precondition for the rest of the test.

## Definition of done
- \`uv run ty check .\` reports 0 diagnostics outside \`.venv\`/stdlib stubs.
- Full test suite green, no behavior change (this is a type-annotation/
  assertion-only change, not a logic change).
- Then wire \`ty check\` into CI (#119) as a hard gate, plus the pre-commit
  config.

neuromechanist:
## Summary

Fixes #122. Drives `uv run ty check .` from **496 diagnostics to 0**
repo-wide, unblocking the CI gate + pre-commit hooks tracked in #119.

The large majority of diagnostics had a single root cause: the stateful
estimators (`AMICATorchNG`, `AMICA_NumPy`, `AMICAMLXNG`) initialize
fit-dependent attributes (`self.W`, `self.A`, `self.sphere`, `self.c`,
`self.comp_list`, ...) to `None` in `__init__` and only assign real
tensors/arrays in `fit()`. `ty` correctly infers these as `Tensor | None`
(etc.) and flags every downstream use.

## Fix approach (no API changes, no blanket silencing)

- **Internal EM-loop call sites:** `assert x is not None` at point of use,
  treating "fit populated this" as an enforced invariant ("trust internal
  code", per CLAUDE.md).
- **Public accessors** (`transform`, `get_mixing_matrix`,
  `get_unmixing_matrix`, `variance_order`, `save`, ...): real
  `RuntimeError` guards, matching the existing `state_dict()`/`_check_usable`
  convention, since a user can call these before `fit()`.
- **Test files:** `assert` right after the relevant `fit()`/helper call,
  which also documents "fit succeeded" as a precondition.
- **Explicit `Optional[...]` annotations** added to the fit-dependent
  attributes so the intent is declared, not just inferred.
- A handful of unrelated real type issues fixed on their own merits:
  `spec_from_file_location(...) -> ModuleSpec | None` narrowing in test
  loaders, `np.dtype` default/annotation mismatches widened to
  `npt.DTypeLike`, `str`-only path params widened to `Union[str, Path]`,
  dict-spread-into-typed-constructor annotated `dict[str, Any]`, and raw
  instance-attribute method monkey-patches replaced with
  `mock.patch.object(...)`.

## Scoped exceptions (documented, not silenced globally)

- `.context/**` excluded from `ty` scope via `[tool.ty.src]` in
  `pyproject.toml`: these are frozen historical investigation scripts for
  closed issues (explicitly marked "do not ship as-is"). Live/regenerated
  analysis scripts were fixed on their own merits, not excluded.
- Two `import mlx.core` lines carry a scoped
  `# ty: ignore[unresolved-import]` with an explanatory comment. `mlx`
  ships as a compiled extension with no `.pyi` stubs, so `ty` cannot
  resolve the import even when the package is installed (verified: the
  diagnostic persists after `uv pip install mlx`). No annotation or assert
  can resolve a missing-stubs import; the suppression is local to those two
  lines only.

## Side-effect finding

While annotating `numpy_impl/core.py` I found `_reject_outliers()`
references `self.data_mask`, which was never initialized anywhere, so
`do_reject=True` would crash. Filed as #123. This PR adds a minimal
non-crashing init (`data_mask` allocated in `_initialize_parameters`) with
a comment noting the feature is still functionally incomplete pending the
real fix in #123 (out of scope here).

## Validation

- `uv run ty check .` -> **All checks passed!** (0 diagnostics)
- Full test suite green (annotation/assert-only change, no logic change)

## Definition of done (#122)

- [x] `uv run ty check .` reports 0 diagnostics
- [x] Full test suite green, no behavior change
- [ ] Wire `ty check` into CI as a hard gate + pre-commit config -> #119


neuromechanist:
## `/review-pr` summary (5 parallel reviewers: code, tests, silent-failure, type-design, comments)

**No Critical findings.** The PR is behavior-preserving; `ty check .` = 0 diagnostics, `ruff` clean, full suite 191 passed / 8 skipped / 3 xfailed / 0 failed. Reviewers surfaced two real gaps (now fixed) plus doc/comment consistency items (now fixed) and two intentional decisions to document (below).

### Fixed in follow-up commits

1. **`do_reject=True` on the NumPy backend crashed mid-fit** (silent-failure HIGH; corroborated by code-review + type-design). `_update_parameters` divides by `self.num_good_samples` every iteration under `do_reject`, but that attribute is only set inside `_reject_outliers` (later, on schedule), so iteration 0 raised a bare `AttributeError` deep in the EM loop, independent of the reject schedule. My initial `data_mask` stub + comment only addressed a *different*, unreached crash and were misleading. **Fix:** `fit()` now refuses `do_reject=True` with a clear `NotImplementedError` pointing at #123 (loud + actionable instead of a cryptic crash or a silent no-op), the misleading comments are corrected, and a regression test (`test_do_reject_refused_on_numpy_backend`) locks the behavior. The real port stays tracked in #123. The PyTorch backend's `do_reject` (via `good_idx`) is unaffected and works.

2. **`AMICATorchNG.write_amica_output` was left unguarded** (silent-failure HIGH). Its four sibling public accessors (`transform`, `get_mixing_matrix`, `get_unmixing_matrix`, `variance_order`) got `RuntimeError` fit-guards in this PR, but `write_amica_output` was missed because its `Optional` attributes flow through an unannotated local `_np` helper that `ty` can't see through. Calling it before `fit()` raised a bare `AttributeError`. **Fix:** added the same `RuntimeError("... requires a fitted model; call fit() first.")` guard.

3. **Doc/comment consistency** (comment + silent-failure reviewers): `docs/api/numpy-backend.md` and `AGENTS.md` claimed the NumPy backend has working "outlier rejection", which now contradicts the honest #123 gate. Both caveated. The `pyproject.toml` `.context/**` comment over-generalized the "do not ship as-is" marking (only one file carries it verbatim); reworded to name that file as the example and note that live/tested `.context` scripts were fixed on their merits, not excluded.

4. **Minor consistency** (test reviewer): `test_ng_sharing.py` now uses lowercase `dict[str, Any]` like every other touched file (was `Dict[str, Any]`); removed the now-redundant `str()` wrapper in `test_amica.py` (`load_data_file` accepts `Path`).

### Intentional decisions (documented per the no-mocks / no-behavior-change policies)

- **`unittest.mock.patch.object` in `test_ng_backend.py`** (code + test reviewers asked to confirm): this is a **spy**, not a mock/fake. `mock.patch.object(ng, "_finalize_newton_stats", side_effect=fn)` where `fn` calls through to the real `_finalize_newton_stats(acc)` — no result is faked, the real EM computation runs. It replaces a raw instance-attribute method reassignment that `ty` correctly flagged as a signature mismatch, and it auto-restores after the `with` block (strictly better than the leaked monkey-patch). Consistent with the "no mock data/tests" policy.
- **`sigma2`/`lambda_`/`kappa` init hoist** (type-design nit): these were previously assigned `None` only inside `if self.do_newton:`; the diff hoists them to unconditional `None` so the `Optional[...]` annotations are honest. This means the attributes now always exist (vs. not existing when `do_newton=False`) — a benign change, no test relies on `hasattr`, and internal call sites are always reached only after `_initialize_parameters` populates the real values.

### Follow-ups (out of scope, tracked)

- #123: real NumPy `do_reject` port (this PR only gates it honestly).
- `_snapshot_params`/`_restore_params` return `Dict[str, object]` (pre-existing, untouched here); a `TypedDict` would remove the `isinstance` narrowing the tests need — worth a separate issue.

PR #125, opened 2026-07-14
CI: enforce ty type-checking + pre-commit config (ruff + ty)

### LINKED ISSUE #119: CI: enforce ty type-checking + add pre-commit config (ruff + ty)

neuromechanist:
## Summary

Closes #119. Enforces `ty` type-checking in CI and adds a `.pre-commit-config.yaml`
with ruff + ty hooks. Unblocked by #122 (repo is now at 0 `ty` diagnostics), so
`ty check` lands as a **hard gate** (option (a) from the issue), not informational.

## Changes

1. **`ty` + `pre-commit` pinned in the dev group** (`pyproject.toml` + `uv.lock`).
   `ty` was only available locally via Homebrew, so CI's ubuntu runner had no type
   checker at all. Pinning both means `uv sync` provides the same versions in CI,
   locally, and in the hook (ty 0.0.59, pre-commit 4.x).

2. **CI `typecheck` job** (`.github/workflows/ci.yml`): a new `Type check (ty)`
   job runs `uv run ty check .` on the plain base env. `test` and `build` now
   `needs: [lint, typecheck]`, so a type error fails fast before the ~10-min test
   run. Added as its own job (not folded into `lint`) so it surfaces as a distinct
   status without renaming the existing `Lint (ruff)` check.

3. **`.pre-commit-config.yaml`**: `local` hooks running ruff + ty **through uv**,
   so versions track pyproject/CI with no pre-commit-mirror drift.
   - ruff `check --fix --unsafe-fixes` + `format` on staged Python files.
   - `ty check .` **whole-project** (`pass_filenames: false`): type errors are
     inherently cross-file, so per-file checking is unsound; ty is fast enough
     (Rust, sub-second here) to re-check everything each commit. Fires only when a
     Python file is staged, and runs the identical command to the CI job, so a
     green pre-commit implies a green CI gate.

4. **Scoped `# ty: ignore[unresolved-import]` on the two lazy `mne` imports** in
   `benchmarks/benchmark_decompose.py`. `mne` is the optional `viz` extra, absent
   from the base env; this matches the existing `mlx` treatment and keeps
   `ty check .` clean on the base env (so every dev and the hook see 0 diagnostics
   without installing optional extras), rather than coupling the gate to `--extra viz`.

## Design notes

- **Why whole-project ty in the hook** (a deliberate deviation from the issue's
  "staged files only" for ty specifically): a type checker that only sees staged
  files misses errors an edit introduces in *other* files. ruff (a fixer) stays
  staged-only where scoping actually matters. The whole-project ty run is cheap.
- **Optional deps stay ignored, not installed**: `mlx` (Apple-only) can never
  resolve on Linux, and `mne` (viz extra) would need a heavier CI env and still
  leave local `ty check .` dirty for anyone without the extra. Scoped inline
  ignores keep the base-env check honestly clean and the gate simple.

## Validation

- `uv run ty check .` -> **All checks passed!** (0 diagnostics, base env)
- `uv run pre-commit run --all-files` -> ruff-check / ruff-format / ty all **Passed**
- `pytest test_decompose_equiv.py` green (only Python change is the two comments)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,9 +46,26 @@ jobs:
       - name: ruff format --check
         run: uv run ruff format --check .
 
+  typecheck:
+    name: Type check (ty)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.12"
+          enable-cache: true
+      # ty is pinned in the dev dependency group, so `uv sync` installs the same
+      # version used locally and by the pre-commit hook. Optional-extra imports
+      # absent from the base env (mlx, mne) carry scoped `# ty: ignore` comments,
+      # so this runs on the plain base env like the other jobs (issue #119).
+      - run: uv sync
+      - name: ty check
+        run: uv run ty check .
+
   test:
     name: Test (Python 3.12)
-    needs: lint
+    needs: [lint, typecheck]
     runs-on: ubuntu-latest
     env:
       PYTORCH_ENABLE_MPS_FALLBACK: "1"
@@ -87,7 +104,7 @@ jobs:
 
   build:
     name: Build + import (Python ${{ matrix.python-version }})
-    needs: lint
+    needs: [lint, typecheck]
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,10 +55,11 @@ jobs:
         with:
           python-version: "3.12"
           enable-cache: true
-      # ty is pinned in the dev dependency group, so `uv sync` installs the same
-      # version used locally and by the pre-commit hook. Optional-extra imports
-      # absent from the base env (mlx, mne) carry scoped `# ty: ignore` comments,
-      # so this runs on the plain base env like the other jobs (issue #119).
+      # ty is in the dev dependency group and version-locked via uv.lock, so
+      # `uv sync` installs the same version used locally and by the pre-commit
+      # hook. Optional-extra imports absent from the base env (mlx, mne) carry
+      # scoped `# ty: ignore` comments, so this runs on the plain base env like
+      # the other jobs (issue #119).
       - run: uv sync
       - name: ty check
         run: uv run ty check .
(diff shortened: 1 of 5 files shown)


neuromechanist:
## /review-pr summary (2 Sonnet reviewers: code + comments)

**No Critical or Important findings.** Both reviewers verified the CI + pre-commit wiring end-to-end (locally ran `uv sync`, `ty check .`, `pre-commit run --all-files`; cross-checked live CI where the new `Type check (ty)` job passed in ~26s and `test`/`build` correctly waited on `needs: [lint, typecheck]`).

Confirmed correct: the acyclic `needs` graph, `paths-ignore` skipping the whole workflow (incl. typecheck) for docs-only changes as before, valid `.pre-commit-config.yaml` (all three hooks execute; `pass_filenames: false` on ty runs whole-project as intended), ty/pre-commit present in `uv.lock`, the mne `ty: ignore` scoping matching the existing mlx treatment, and UV-only / no-emoji / no-AI-attribution compliance. Comment reviewer verified every new comment is factually accurate (mne = viz extra, mlx = Apple-only, hook `entry` byte-identical to the CI step, #119 is the right issue, sub-second ty runtime measured).

### Addressed (minor precision nits, commit 1065499)
- Reworded 'ty is pinned in the dev dependency group' -> 'in the dev dependency group and version-locked via uv.lock' (pyproject uses a lower bound; the lockfile does the exact pinning) in both the CI comment and the hook header.
- Install instruction now notes the `uv sync` prerequisite: 'Install once (after `uv sync`): uv run pre-commit install'.

No other changes needed.

PR #126, opened 2026-07-14
Implement NumPy do_reject outlier rejection via good_idx (#123)

### LINKED ISSUE #123: numpy_impl: do_reject outlier rejection is non-functional (data_mask never initialized/consumed)

neuromechanist:
## Summary

Closes #123. Implements outlier rejection (`do_reject`) on the NumPy backend by
porting `AMICATorchNG`'s `good_idx` mechanism. The feature was previously dead:
`_reject_outliers` thresholded the *scalar* total LL (not per-sample),
`self.data_mask` was never consumed by the E-step, and `num_good_samples` was
used before assignment (`do_reject=True` crashed mid-fit; #124 gated it with a
temporary `NotImplementedError`).

## Changes

- **`good_idx`**: an index array of the currently-kept samples, initialized to
  all indices in `_initialize_parameters` and only ever shrinking, with
  `num_good_samples` tracking its size (mirrors the torch backend exactly).
- **`_get_updates_and_likelihood`**: restricts the E-step to `data[:, good_idx]`
  under `do_reject`, and captures the pre-update per-sample LL (in `good_idx`
  order) so the reject keep-mask maps straight back onto `good_idx`.
- **`_get_block_updates`**: exposes the per-sample LL vector (`ll_samples`) under
  `do_reject`; `updates["ll"]` is `np.sum` of the same vector (unchanged).
- **`_reject_outliers`**: thresholds that per-sample LL at `mean - rejsig*std`
  (population std), shrinks `good_idx`, and raises a clear `ValueError` if every
  sample would drop -- byte-for-byte the torch backend's logic.
- **`_update_parameters`**: `gm` normalizes by `num_good_samples` under
  `do_reject` (the count the E-step actually summed over).
- **Construction-time validation** of `rejint`/`rejsig`/`maxrej` (matching
  `AMICATorchNG`), so `rejint=0` / `rejsig<=0` fail up front instead of as an
  opaque `ZeroDivisionError` / empty-good-set crash deep in the loop.
- Removes the #124 `NotImplementedError` gate and the dead `data_mask` attribute.

## `do_reject=False` is bit-identical

Every addition is guarded by `if self.do_reject`, and `data_use` is `self.data`
(no copy) on the default path; the `ll` computation is `np.sum` of the exact same
expression as before. So the issue #24 single-model parity path is unchanged.

## Tests (`pyAMICA/tests/test_numpy_reject.py`, real sample EEG -- no synthetic)

- `test_rejection_shrinks_good_sample_set`: good set shrinks, `good_idx` stays a
  unique in-range subset, `num_good_samples` tracks it, W/LL finite.
- `test_gm_normalizes_by_good_count_after_rejection`: multi-model (`n_models=2`),
  `gm` stays a valid finite distribution summing to 1 after rejection.
- `test_rejection_degenerate_ll_raises_clear_error`: all-NaN per-sample LL ->
  `ValueError("removed all ...")`.
- `test_reject_param_validation`: `rejint=0` / `rejsig<=0` raise at construction.
- `test_do_reject_false_leaves_good_idx_unset`: default path never touches the
  machinery (`good_idx is None`).
- Drops the now-obsolete `test_do_reject_refused_on_numpy_backend` (the #124
  refusal test).

## Validation

- `uv run ty check .` -> All checks passed! (0 diagnostics)
- New + existing NumPy tests green (`test_numpy_reject.py`, `test_pyAMICA.py`)
- Parity regression: `test_amica.py` + `test_sample_data.py` -> 10 passed,
  3 xfailed (pre-existing), 0 failed -- default path unregressed.

## Note on parity

There is no bit-exact Fortran `do_reject` oracle (the torch backend's rejection
is itself behavior-validated, not bit-exact), so this is validated the same way:
the mechanism matches the torch port and behaves correctly on real data.


neuromechanist:
## /review-pr summary (4 Sonnet reviewers: code, tests, silent-failure, comments)

Reviewers found real issues (2 verified experimentally against live `AMICA_NumPy`/`AMICATorchNG`). All addressed in commit `763b971`.

### Critical (fixed)
1. **Reject schedule missing the `max(1, iter - rejstart)` clamp** (code review, verified). Python's non-negative modulo let `(iter - rejstart) % rejint == 0` fire rejection *before* `rejstart` for non-default settings (e.g. `rejstart=3, rejint=2` fired at [1,3,5] instead of [3,5]). This pre-existing schedule code was inert while do_reject was non-functional; my PR activated it. Added the clamp matching Fortran (amica17.f90:1142) + torch; verified it now fires at [3,5,7].
2. **Misleading LL-normalization comments** (silent-failure + comment review, both independently). My `good_idx` comments claimed `num_good_samples` drives the *gm/LL* normalization, but `self.ll` is a raw sum, never normalized (this backend's convention for both modes). Corrected to say only `gm` is normalized, and documented that dropping the most-negative samples *raises* the raw-sum LL, so the reject iteration is an increase and does not spuriously trip convergence. **Chose not to normalize** self.ll: the `do_reject=False` path must stay bit-identical for #24 parity (normalizing would shift `min_dll` timing). Added `test_rejection_does_not_spuriously_stop_fit` to lock the behavior.

### Important (fixed)
3. **Restart reverts prior rejections** (code review). `_reinitialize_for_restart` reset `good_idx` to the full set; Fortran's startover preserves rejections (amica17.f90:1121-1148). Now snapshots/restores `good_idx`/`num_good_samples` across the reinit.
4. **`numrej` was a local, maxrej cap untested** (test review). Exposed `self.numrej` (matching torch, testable) and asserted the cap.
5. **Multi-model `c` finiteness untested** (test review). Added `test_multimodel_rejection_keeps_gm_and_c_finite` exercising the dead-model `dgm==0` guard.

### Minor (fixed)
- `rejsig<=0` validation wording corrected (at 0 it keeps ~half, not "every sample"; comment review); added `rejstart>=0` validation; docstrings made accurate ("exactly maxrej passes" now asserted).

### Follow-up (filed, not this PR)
- **#127**: both backends' `_reject_outliers` raise a `rejsig`-blaming message when the real cause is a non-finite per-sample LL (numerical instability). Pre-existing in torch, faithfully ported; fix both together to avoid divergence.

Validation: `ty check .` 0 diagnostics, ruff clean, all 6 do_reject tests pass, `do_reject=False` still bit-identical (#24 parity path untouched).

PR #129, opened 2026-07-14
Reject: distinguish non-finite LL from too-aggressive rejsig (#127)

### LINKED ISSUE #127: Outlier rejection: distinguish non-finite per-sample LL from a too-aggressive rejsig
Found in the PR #126 review (silent-failure). In both backends' `_reject_outliers`, a single non-finite per-sample log-likelihood (e.g. a sample driving `logdet_W` or an activation to +/-inf from numerical instability) makes `mean`/`std` NaN, so `keep = ll >= mean - rejsig*std` is all-False and the guard raises:

> Outlier rejection removed all N samples (rejsig=X too aggressive for this data).

That message blames the `rejsig` hyperparameter when the real cause is upstream numerical instability; a user raising `rejsig` in response won't fix anything.

## Scope
Detect non-finite entries in the per-sample LL explicitly and raise a distinct, accurate error (e.g. "K samples have non-finite log-likelihood, indicating numerical instability (singular W / overflow), not a rejsig miscalibration") before the all-rejected check.

**Fix both backends together** to avoid cross-backend divergence:
- `pyAMICA/torch_impl/core.py` `AMICATorchNG._reject_outliers` (pre-existing; `ll_vec.mean()` with no NaN guard).
- `pyAMICA/numpy_impl/core.py` `AMICA_NumPy._reject_outliers` (same pattern, ported in #123).

Not a regression from #123; the NumPy port faithfully reproduced the existing torch behavior. Low severity (needs a genuinely non-finite per-sample LL to trigger).

neuromechanist:
## Summary

Closes #127. Makes both backends' `_reject_outliers` report the real cause when
outlier rejection would drop every sample.

Builds on #126 (now merged); rebased onto `main`.

## Problem

Both `AMICATorchNG._reject_outliers` and `AMICA_NumPy._reject_outliers` raised:

> Outlier rejection removed all N samples (rejsig=X too aggressive for this data).

whenever the keep-mask was empty. But for **finite** log-likelihoods with the
construction-validated `rejsig > 0`, the maximum sample is always `>= mean >=
mean - rejsig*std`, so it is always kept -- the good set can only empty when the
per-sample LL is **non-finite** (a single NaN poisons `mean`/`std`, making every
comparison `False`). That is upstream numerical instability (singular `W` /
overflow), which a user cannot fix by tuning `rejsig`; the old message
misdiagnosed it.

## Change

On an empty keep-mask, both backends now count the non-finite entries and raise
an accurate instability message naming the real cause. The old `rejsig` message
is kept only as a defensive backstop (now unreachable for finite LL with
`rejsig > 0`). The two backends raise the same message style, fixed together to
avoid cross-backend divergence.

## Tests

Both backends' degenerate-LL tests now inject a **single** NaN (demonstrating one
non-finite entry poisons the whole reject decision) and assert the `non-finite`
message:
- `test_numpy_reject.py::test_rejection_nonfinite_ll_raises_instability_error`
- `test_ng_backend.py::test_rejection_degenerate_ll_raises_clear_error`

## Validation

- `uv run ty check .` -> 0 diagnostics; ruff clean.
- All reject tests pass (6 NumPy + 3 torch).

### DIFF
--- a/pyAMICA/numpy_impl/core.py
+++ b/pyAMICA/numpy_impl/core.py
@@ -1465,7 +1465,21 @@ class AMICA:
         keep = ll_vec >= (mean - self.rejsig * std)
 
         if not bool(keep.any()):
-            raise ValueError(
+            # For finite log-likelihoods the max sample is always >= mean >=
+            # mean - rejsig*std (rejsig>0 is validated at construction), so it is
+            # always kept; the only way every sample is dropped is a non-finite
+            # per-sample LL (one NaN poisons mean/std, making every comparison
+            # False). Report that accurately instead of blaming rejsig (issue
+            # #127), which a user cannot fix by tuning rejsig.
+            n_bad = int(np.count_nonzero(~np.isfinite(ll_vec)))
+            if n_bad:
+                raise ValueError(
+                    f"{n_bad} of {ll_vec.size} samples have a non-finite "
+                    "log-likelihood; this indicates numerical instability "
+                    "(singular W / overflow) upstream, not a rejsig "
+                    "miscalibration."
+                )
+            raise ValueError(  # defensive: unreachable for finite LL, rejsig>0
                 f"Outlier rejection removed all {self.good_idx.size} samples "
                 f"(rejsig={self.rejsig} too aggressive for this data)."
             )
--- a/pyAMICA/numpy_impl/core.py
+++ b/pyAMICA/numpy_impl/core.py
@@ -1470,14 +1470,18 @@ class AMICA:
             # always kept; the only way every sample is dropped is a non-finite
             # per-sample LL (one NaN poisons mean/std, making every comparison
             # False). Report that accurately instead of blaming rejsig (issue
-            # #127), which a user cannot fix by tuning rejsig.
+            # #127), which a user cannot fix by tuning rejsig. In a normal fit()
+            # the earlier aggregate non-finite-LL guard (the sum is non-finite
+            # iff a term is) stops the loop first, so this mainly serves direct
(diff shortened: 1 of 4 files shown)


neuromechanist:
## /review-pr summary (2 Sonnet reviewers: code + silent-failure)

**No Critical/Important correctness issues** in either review; both verified the change end-to-end (including +inf/overflow edge cases, single-NaN poisoning, raise-before-mutate, cross-backend consistency, and that the backstop `raise` is genuinely dead code for finite LL with rejsig>0). Two points addressed in commit `8f9bbdc`:

- **Scope (silent-failure Medium)**: the new accurate branch is effectively unreachable from a real `fit()` -- both backends' earlier aggregate non-finite-LL guard (the LL sum is non-finite iff a per-sample term is) stops the loop first, so users see `nan_ll`/`singular_ll` (torch) or the non-finite convergence stop (numpy), not this message. So this fix is defense-in-depth for direct callers of `_reject_outliers`, not an observed-in-prod change. Noted explicitly in the code comment (and the PR framing).
- **Actionable message (Low)**: the message now names a likely remedy (rank-deficient/average-referenced data, lower learning rate), matching the sibling singular-W message's house style, rather than only stating what is *not* wrong.

Not changed (accepted): the mismatched test names (code review, confidence ~30, 'not worth a revision') and the theoretical astronomically-large-finite-LL overflow gap (silent-failure Low, not physically reachable in this pipeline).

Validation: `ty check .` 0 diagnostics, ruff clean, all reject tests pass (6 NumPy + torch degenerate).

Issue #134, opened 2026-07-14
Phase 1: MIR core metric (mutual information reduction)


neuromechanist:
Part of #133.

Port `mir(unmixing, data) -> (mir_nats, variance)` from `bigdelys/pre_ICA_cleaning`'s
`getMIR.m` (Apache-2.0, permits direct port with attribution) using pyAMICA's
existing sphere + unmixing pipeline.

- Use Seyed Yahya Shirazi's own `mir()` adaptation in
  `sccn/NEMAR-pipeline/eeg_nemar_dataqual.m` as the reference (it adds the
  sphering step), but fix the scope bug: that version's signature is
  `(data, linT)` yet its body references `eig(W)` and `/N`, which only resolve
  today via MATLAB's nested-function shared workspace. The Python port must use
  explicit parameters (`linT`, `N = data.shape[1]`), not replicate the implicit
  dependency.
- Add an attribution docstring/comment pointing to
  `github.com/bigdelys/pre_ICA_cleaning/getMIR.m` per the Apache-2.0 NOTICE
  requirement.
- Tests: real bundled sample EEG only (NO MOCK data), checking internal
  consistency (e.g. against a literal re-derivation of the marginal-entropy
  terms), not synthetic data.

neuromechanist:
Done in PR #153, squash-merged into the epic branch `feature/issue-133-epic-mir-pmi-metrics` (e2ff460). Closing manually: the PR targeted the epic branch rather than the default branch, so GitHub's auto-close on merge did not fire. Ships to main with epic #133.

Issue #137, opened 2026-07-14
Phase 5: wire MIR/PMI into the AMICA API


neuromechanist:
Part of #133. Depends on Phases 1-3.

Expose `model.mir(X)` / `model.pmi()` on `AMICA` and `AMICATorchNG`, computable
at fit-end and optionally at specified iteration waypoints, per Scott Makeig's
original request (applicable "at decomposition end and/or optionally at
specified decomposition waypoints").

neuromechanist:
Done in PR #160, squash-merged into the epic branch `feature/issue-133-epic-mir-pmi-metrics` (4e47885). Closing manually: the PR targeted the epic branch rather than the default branch, so GitHub's auto-close on merge did not fire. Ships to main with epic #133.

`model.mir(X)` / `model.pmi(X)` on both `AMICA` and `AMICATorchNG`, plus a `mir_step` waypoint schedule recording a MIR trajectory during the fit (MIR only: measured at 20ms vs PMI's 678ms against a 36-400ms fit iteration, so PMI on a schedule would dominate the fit for a quantity that is a final diagnostic rather than a convergence signal).

Review caught a real bug: a failed waypoint propagated out of `fit()` and destroyed the whole decomposition. Now warns and records a visible NaN gap. Follow-up #161 tracks two documented-but-untested claims about `mir_history_`.

Issue #133, opened 2026-07-14
Epic: port MIR and PMI separation-quality metrics from postAmicaUtility/pre_ICA_cleaning


neuromechanist:
## Background

Reviewer feedback on the JOSS paper: Scott Makeig strongly recommended adding
Mutual Information Reduction (MIR, Palmer's separation-quality metric from
Delorme et al. 2012) as a built-in pyAMICA option, applicable at decomposition
end and/or at specified waypoints. Investigating where MIR and the related PMI
metric actually live turned up two distinct upstream sources with different
licenses -- important since pyAMICA is BSD-3-Clause:

- **MIR**: `getMIR.m` in
  [bigdelys/pre_ICA_cleaning](https://github.com/bigdelys/pre_ICA_cleaning)
  (Arnaud Delorme's repo) -- **Apache License 2.0** (permissive, attribution
  required). Seyed Yahya Shirazi already adapted this into a `mir()` nested
  function in `sccn/NEMAR-pipeline/eeg_nemar_dataqual.m` (2023), with a sphering
  step added ("sphereing is needed to make sure that the MIR is only related to
  ICA"). That adapted version has a latent bug worth *not* reproducing: `mir()`'s
  signature is `(data, linT)` but its body references `eig(W)` and `/N`, neither
  of which are parameters -- it only runs today because MATLAB nested functions
  share the parent function's workspace and happen to have same-named variables
  in scope. The Python port should use explicit parameters (`linT`, `N =
  data.shape[1]`) throughout, not replicate the implicit-scope dependency.

- **PMI** (Pairwise Mutual Information, a related but distinct metric -- NOT the
  same as MIR): lives in
  [sccn/postAmicaUtility](https://github.com/sccn/postAmicaUtility)
  (`minfojp.m`/`get_mi.m`/`arrminf2.m`/`pop_modPMI.m`), a MATLAB/EEGLAB plugin,
  last touched 2023, GPL-2.0-or-later per file headers (no repo LICENSE file).
  There is no MIR implementation anywhere in postAmicaUtility (grepped for
  "mutual information reduction" and `\bMIR\b`, zero matches) -- the earlier
  assumption that MIR lived there was wrong.

## Licensing approach (decided)

- MIR: Apache-2.0 permits a direct port; carry an attribution notice to
  `bigdelys/pre_ICA_cleaning` in the module docstring per the Apache-2.0
  attribution requirement.
[...]

PR #153, opened 2026-07-16
Phase 1: MIR core metric (mutual information reduction)

neuromechanist:
## Summary

Ports the MIR (Mutual Information Reduction) separation-quality metric from `getMIR.m` in `bigdelys/pre_ICA_cleaning` (Apache-2.0, permits a direct port with attribution). Adds `pyAMICA.metrics.mir(unmixing, data, nbins=None) -> (mir_nats, variance)` as a standalone, backend-agnostic function.

The `_marginal_entropies` helper (a port of `getMIR.m`'s nested `getent4`) uses `np.unique(binned, return_counts=True)` rather than `np.histogram`, since the MATLAB source's run-length-of-sorted-values method only ever reports occupied bins, which is what keeps `log(p)` finite.

This does not port the 2023 NEMAR-pipeline `mir(data, linT)` adaptation directly, since that version has a real scope bug (its body references `eig(W)` and `/N`, neither of which are its own parameters, and only resolves via MATLAB's nested-function shared workspace) and depends on a GPL-licensed sphering routine out of scope for this BSD-3-Clause project. Instead, `mir()` takes an explicit `unmixing` matrix; composing pyAMICA's own sphere-then-unmixing pipeline into that matrix is deferred to the API-wiring phase (#137).

Not included in this phase: wiring `mir()` into the `AMICA`/`AMICATorchNG` classes (`model.mir(X)`) -- that's phase 4, issue #137.

Closes #134
Part of epic #133

## Test plan

- `pyAMICA/tests/test_metrics_mir.py` (new): 5 tests against the real bundled sample EEG data (no synthetic/mock data), including an exact closed-form check (scaled-identity unmixing gives MIR ~= 0 and a provably-exact variance) and an independent `np.histogram`-based re-derivation of the entropy estimator.
- `uv run pytest pyAMICA/tests/test_metrics_mir.py -v`: 5 passed.
- `uv run pytest pyAMICA/tests/ -q -m "not slow"`: 191 passed, 9 skipped, 6 deselected, 1 xfailed (pre-existing) -- no regressions.
- `uv run ruff check` / `uv run ruff format` / `uv run ty check`: all clean.

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -1,0 +1,196 @@
+# Third-Party Notices
+
+pyAMICA is BSD-3-Clause (see `LICENSE`). This file lists source code ported
+into the package from third-party projects under other licenses, and
+reproduces the license text required by each, per that license's
+redistribution terms.
+
+## MIR (Mutual Information Reduction) metric
+
+`pyAMICA/metrics/mir.py` is ported from `getMIR.m` (and its nested
+`getent4`) in Arnaud Delorme's
+[`bigdelys/pre_ICA_cleaning`](https://github.com/bigdelys/pre_ICA_cleaning)
+repository, at
+[commit `b6034f0`](https://github.com/bigdelys/pre_ICA_cleaning/blob/b6034f03889a6a418968ee119123f3df55251957/getMIR.m),
+licensed under the Apache License, Version 2.0. The upstream file carries no
+embedded copyright header of its own. The full Apache-2.0 license text is
+reproduced below, as required by Section 4(a) of that license.
+
+---
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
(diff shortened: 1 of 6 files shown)


neuromechanist:
## Review summary (4 parallel Sonnet reviewers: code, tests, comments, silent-failures)

**code-reviewer**: no issues; confirmed algorithmic parity against the actual upstream getMIR.m/getent4 line-by-line, including the subtle variance-before-correction statement order.

**silent-failure-hunter** (3 critical, empirically verified by executing the code):
- Singular/near-singular unmixing silently returned a wrong-but-finite MIR value (floating-point noise means eigvals rarely hits exact zero, so log|eigenvalue| stayed finite). Fixed: `mir()` now checks the smallest |eigenvalue| and raises `ValueError` below a 1e-10 threshold, matching this repo's existing logdet-singularity guard pattern in `numpy_impl/core.py`.
- Constant channel silently produced `nan` in the aggregate output (defeats any downstream `mir_val < threshold` QC gate). Fixed: `_marginal_entropies` now raises `ValueError` naming the offending channel.
- NaN/Inf in `data` silently propagated to `nan` with zero warnings. Fixed: `_marginal_entropies` now validates finiteness up front.
- (Medium) `nbins` auto-formula could hit 0 for tiny sample counts; now validated and raises.

**pr-test-analyzer** (1 critical + several important):
- No test exercised a real, non-trivial unmixing matrix (identity/scaled-identity can only ever give MIR ~= 0). Added `test_mir_real_fitted_unmixing_is_large_and_positive`, using the bundled real fitted `pyresults/W.npy`/`sphere.npy` -- confirms MIR ~= 3.8 for the real fit vs exactly 0 for identity.
- Added: variance cross-check in the histogram-rederivation test, an `nbins`-passthrough test at the public `mir()` level, and tests locking in the four new error paths above (plus non-square unmixing, which already failed loudly via `LinAlgError` and now has a test confirming that contract).
- Tightened the identity-unmixing tolerance to match the scaled-identity test (both are the same closed form) and added the missing variance assertion there.

**comment-analyzer** (1 critical + several precision fixes):
- Apache-2.0 attribution named the license but didn't satisfy Section 4(a) (must give recipients a copy of the license itself). Added `THIRD_PARTY_NOTICES.md` vendoring the full Apache-2.0 text, and pinned the `getMIR.m` link to the exact commit it was ported from.
- Softened the overgeneral docstring claim ("an orthogonal transform gives MIR near zero" -> qualified to identity/scalar multiples only, since a generic rotation of dependent non-Gaussian data need not).
- Fixed two wording-precision issues: the dead NEMAR `mir()` code now correctly says it would raise an undefined-variable error if uncommented (not just "resolves via shared workspace"), and the GPL attribution now points at the actual GPL-licensed helper functions it calls (`block_geometric_median`, `hlp_memfree`) rather than `robust_sphering_matrix` itself.

Tests: 11/11 new (up from 5), full suite 197 passed / 9 skipped / 6 deselected / 1 xfailed (pre-existing) -- no regressions. ruff/ty clean.

Issue #155, opened 2026-07-16
Phase 3: posterior-probability (LLt) output parity with Fortran


neuromechanist:
Part of #133. Blocks #136 (visualizations need real multi-model posterior data).

## Background

While planning Phase 3 (MIR/PMI visualizations, #136 -- becoming Phase 4), the multi-model posterior-probability timeline plot needed real per-timepoint, per-model posterior data. This turned up a genuine parity gap, not just a missing input for the viz phase.

Fortran writes an `LLt` file every fit (`amica15.f90`/`amica17.f90` `write_output`, gated by `write_LLt`, default `.true.`): a per-timepoint, per-model log-likelihood array (`modloglik`) plus the total per-sample log-likelihood (`loglik`), natural log, laid out as `[model_1..model_M, total]` per timepoint, sequential raw float64. It's a fresh E-step snapshot each time `write_output` runs (every `writestep` iterations, and once more after the loop), not an accumulated history.

Neither pyAMICA backend writes this file. `_write_results` (`pyAMICA/numpy_impl/core.py`) and `write_amica_output` (`pyAMICA/torch_impl/core.py`) both delegate to `pyAMICA/numpy_impl/load.py`'s `write_amicaout()`, which writes `gm/A/W/S/mean/c/alpha/mu/sbeta/rho/comp_list/LL` -- no `LLt`.

Separately, the existing reader has a confirmed, latent bug. `loadmodout()` (`pyAMICA/numpy_impl/load.py`) reshapes the raw `LLt` bytes with `order='C'` (numpy's default); the file's actual layout is column-major w.r.t. a `(num_models+1, N)` matrix (matching `pyAMICA/sample_data/loadmodout15.m`'s MATLAB reader, column-major by default there). Verified empirically against the real bundled `pyAMICA/sample_data/amicaout/LLt` fixture (num_models=1, so `Lht` must equal `Lt` exactly): current `order='C'` gives `Lht != Lt` for 87% of samples; `order='F'` gives exact equality everywhere, and `mean(Lt) ~= -108.86` matches `nw * final_ll = 32 * -3.40187 ~= -108.86` independently. This bug is latent today only because nothing exercises `Lht`/`Lt`/`v` anywhere in the repo (`validate_implementations.py` never calls `loadmodout`; no test reads these fields).

## Scope

1. Fix the reader's reshape bug (`order="F"`), with a regression test against the real bundled `amicaout/LLt` fixture.
[...]

PR #156, opened 2026-07-16
Phase 3: posterior-probability (LLt) output parity with Fortran

neuromechanist:
## Summary

The reference Fortran binary writes an `LLt` file on every run: the per-timepoint, per-model log-likelihood (`modloglik`) plus the total per-sample log-likelihood (`loglik`). Neither pyAMICA backend produced this file, and the reader had a separate latent bug. Since parity with the Fortran reference is this project's definition of correctness, a reference artifact we fail to reproduce is a regression, not a missing nice-to-have. This phase closes both gaps.

Two distinct bugs:

1. **Missing writer.** `write_amicaout()` emitted `gm/A/W/S/mean/c/alpha/mu/sbeta/rho/comp_list/LL` but never `LLt`, so no pyAMICA run produced the file at all.

2. **Reader reshape order (pre-existing, latent).** `loadmodout()` reshaped the raw `LLt` bytes with numpy's default `order="C"`, but Fortran's `write_output` (amica15.f90:2308-2333) writes, per timepoint, each model's log-likelihood then the total, which is a column-major `(num_models+1, N)` layout (matching EEGLAB's `loadmodout15.m`). Verified against the bundled `sample_data/amicaout/LLt` fixture: under `order="C"` the single-model `Lht` disagrees with `Lt` for 87% of samples, where they must be exactly equal; under `order="F"` they match exactly and `Lt.mean()` recovers `nw * final_ll` independently. This was latent only because nothing in the repo read `Lht`/`Lt`/`v`.

## Approach

Neither backend retains a full-dataset per-sample log-likelihood after fit; it is computed per block during the E-step and discarded once folded into sufficient statistics. Both backends therefore recompute it with a fresh full-dataset forward pass at write time, from whatever parameters are currently on the model. This is what makes the output correct after a `keep_best` rollback (#51), where the final exported parameters are an earlier iterate than the training loop's last.

`modloglik` is confirmed against the Fortran as `log|det W_h| + log(gm_h) + sldet + sum_i ll_src`, and `loglik` as its log-sum-exp over models.

Closes #155
Part of epic #133

## Test plan

All against real bundled sample EEG and the real Fortran-produced fixture; no synthetic data.

- [x] `test_loadmodout_llt_reshape_order` reads the real Fortran `amicaout/LLt` fixture and asserts single-model `Lht[0] == Lt` exactly plus `Lt.mean() ~= nw * final_ll`. This is the oracle that pins the layout to Fortran; it fails under the old `order="C"`.
- [x] `test_numpy_llt_roundtrip` and `test_write_amica_output_llt_roundtrip` fit real single-model NumPy/NG models and round-trip the written `LLt` through the fixed reader. A writer/reader order mismatch scrambles the interleave and breaks these, so they transitively pin the writer to Fortran's layout.
- [x] `test_write_amica_output_llt_multimodel` checks a real 2-model NG fit satisfies `Lt == logsumexp(Lht, axis=0)`.
- [x] Full suite green on the rebased branch: 218 passed, 9 skipped (MLX absent), 1 pre-existing xfail.
- [x] `ruff check`, `ruff format --check`, `ty check` all clean.

### DIFF
--- a/pyAMICA/amica.py
+++ b/pyAMICA/amica.py
@@ -362,6 +362,11 @@ class AMICA:
         sign-flipping is needed. Single-model output is byte-compatible with the
         Fortran reference (issue #92).
 
+        Also writes ``LLt`` (the per-sample/per-model log-likelihood, issue
+        #155) for a model that was just fit in this process; a model restored
+        via :meth:`load` has no training data to recompute it from, so
+        ``LLt`` is omitted for it (a warning is logged).
+
         Parameters
         ----------
         outdir : str
--- a/pyAMICA/numpy_impl/core.py
+++ b/pyAMICA/numpy_impl/core.py
@@ -953,6 +953,86 @@ class AMICA:
 
         return updates
 
+    def _compute_full_posterior_ll(
+        self, data: Optional[np.ndarray] = None
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        """Recompute the per-model/per-sample log-likelihood over every sample
+        of ``data`` (default ``self.data``), for the Fortran ``LLt`` output
+        (issue #155).
+
+        Duplicates the ``b``/``z``/``logV`` forward pass in
+        ``_get_block_updates`` (same definitions, Fortran amica17.f90:1280-1372)
+        rather than sharing it, so this read-only computation cannot affect
+        that method's training-path behavior. Computed fresh from ``self``'s
+        current parameters, not stashed mid-training-loop values, so it is
+        correct even after a keep-best rollback (issue #51).
+
+        Returns
+        -------
+        Lht : ndarray of shape (num_models, n_samples)
+            Per-model log-likelihood, Fortran's ``modloglik``.
+        Lt : ndarray of shape (n_samples,)
+            Total log-likelihood, Fortran's ``loglik``.
+        """
+        assert (
+            self.data_dim is not None
+            and self.W is not None
+            and self.c is not None
+            and self.comp_list is not None
+            and self.beta is not None
+            and self.mu is not None
+            and self.rho is not None
+            and self.alpha is not None
+            and self.gm is not None
+        )
+        if data is None:
+            assert self.data is not None
+            data = self.data
+        n_samples = data.shape[1]
+        Lht = np.zeros((self.num_models, n_samples))
+        Lt = np.zeros(n_samples)
+
+        for start in range(0, n_samples, self.block_size):
+            end = min(start + self.block_size, n_samples)
+            X = data[:, start:end]
+            batch_size = X.shape[1]
+
+            b = np.zeros((batch_size, self.data_dim, self.num_models))
+            for h in range(self.num_models):
+                b[:, :, h] = np.dot((X - self.c[:, h][:, None]).T, self.W[:, :, h])
+
+            z = np.zeros((batch_size, self.data_dim, self.num_mix, self.num_models))
+            for h in range(self.num_models):
+                for i in range(self.data_dim):
+                    k = self.comp_list[i, h]
+                    for j in range(self.num_mix):
+                        y = self.beta[j, k] * (b[:, i, h] - self.mu[j, k])
+                        log_pdf, _ = self._compute_log_pdf(y, self.rho[j, k])
+                        z[:, i, j, h] = (
+                            np.log(self.alpha[j, k]) + np.log(self.beta[j, k]) + log_pdf
+                        )
+
+            z0max = np.max(z, axis=2, keepdims=True)
+            ll_src = z0max[:, :, 0, :] + np.log(np.sum(np.exp(z - z0max), axis=2))
+            logV = np.zeros((batch_size, self.num_models))
+            for h in range(self.num_models):
+                with np.errstate(divide="ignore", over="ignore", invalid="ignore"):
+                    _, logdet_W = np.linalg.slogdet(self.W[:, :, h])
+                logV[:, h] = (
+                    np.log(self.gm[h])
+                    + logdet_W
+                    + self.sldet
+                    + np.sum(ll_src[:, :, h], axis=1)
+                )
+
+            Vmax = np.max(logV, axis=1, keepdims=True)
+            ll_samples = Vmax[:, 0] + np.log(np.sum(np.exp(logV - Vmax), axis=1))
+
+            Lht[:, start:end] = logV.T
+            Lt[start:end] = ll_samples
+
+        return Lht, Lt
+
     def _update_parameters(self, updates: Dict):
         """
         Update model parameters using computed updates.
@@ -1521,6 +1601,8 @@ class AMICA:
         # (loadmodout treats 'nd' as optional).
         from .load import write_amicaout
 
+        Lht, Lt = self._compute_full_posterior_ll()
+
         write_amicaout(
             self.outdir,
             gm=self.gm,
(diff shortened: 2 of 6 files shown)


neuromechanist:
## Review summary

Reviewed by four parallel agents (code, tests, comments, silent failures) plus a maintainer pass against the Fortran reference. CI was green on the first push, but the review surfaced a real parity bug. All findings below are being addressed in follow-up commits; none are being deferred.

### Critical

**1. `do_reject` breaks LLt parity with Fortran.** Found independently by the test reviewer, the silent-failure reviewer, and the maintainer pass. Fortran zeroes `modloglik(:,i)` and `loglik(i)` for rejected samples before writing (`amica15.f90:2211-2216`). Those zeros are load-bearing, not incidental: Fortran's `load_rej` reads `LLt` back and reconstructs the rejection mask by testing `sum(modloglik(:,i)) == 0.0` as the sentinel for a rejected sample (`amica15.f90:887-896`). Both backends currently compute over the full unfiltered dataset and write real log-likelihoods everywhere, so a Fortran `load_rej` run reading a pyAMICA `LLt` would silently conclude nothing was rejected. Secondary hazard: rejected outlier samples were being pushed through the model for the first time at write time. Fix: compute only over `good_idx` and scatter into zero-filled full-length arrays, so rejected samples are exactly `0.0` and are never evaluated. Guarded by new tests on both backends.

**2. Restored models silently write no LLt.** `fit()` -> `save()` -> `load()` -> `write_amica_output()` is a fully public path that produces an output directory missing `LLt`, with no warning at any layer. `from_state_dict` never sets the training data, and the wrapper's `_check_usable` does not know about it. The wrapper docstring does not even list `LLt` among the written files, so nothing documented was being violated either. Fix: explicit `logger.warning` naming the restored-model case, plus docstring updates on both the backend and the wrapper.

### Important

**3. `_X_t` retained the full sphered dataset for the model's lifetime.** The torch backend previously freed the training data after `fit()`; retaining it pins `n_channels * N * 8` bytes, on GPU too, for exactly the large-data case the backend exists to serve. Fix: compute LLt once at the end of `fit()` (strictly after any `keep_best` rollback, so it reflects the exported parameters) and retain only the compact `(n_models+1, N)` arrays, roughly 16x smaller at 32 channels. The NumPy backend is unchanged here: it already retained `self.data` before this PR and must keep writing at mid-training `writestep` boundaries.

**4. NumPy duplicated the forward pass instead of sharing it.** The torch side correctly reuses `_forward`; the NumPy `_compute_full_posterior_ll` was a hand copy of `_get_block_updates`'s b/z/logV math. This repo has had one-sided forward-pass bugs before (#24's fp-vs-dpdf sign error), and a future fix to one copy would silently produce plausible-but-wrong LLt from the other. The copy had also dropped the `isfinite(logdet_W)` guard the original carries, so a singular W would write silent `-inf`/`NaN` into LLt. Fix: extract one shared `_forward_block` used by both callers, which restores the guard on the write path. Being verified bit-exactly against the pre-refactor LL trajectory, since this touches the training path.

**5. `write_amicaout` silently skipped LLt when given only one of `Lht`/`Lt`.** Not reachable today (both call sites pass them as a computed pair) but a latent footgun in a public writer. Fix: raise `ValueError` on a half-specified pair.

### Test gaps being closed
`do_reject` + LLt on both backends; NumPy multi-model LLt (only torch had one, and the two implementations are independent); `from_state_dict` -> `write_amica_output`; and a partial final block (all three original tests used sample counts exactly divisible by `block_size`, so the remainder branch was never exercised).

### Verified and cleared
- `test_loadmodout_llt_reshape_order` is a genuine regression guard, confirmed empirically: reverting to the old `order="C"` gives a max abs difference of ~58.7 between `Lht[0]` and `Lt` on the real bundled fixture, where they must be exactly equal.
- The writer is transitively pinned to Fortran's layout: the fixture test pins the reader to `order="F"`, and the round-trip tests pin writer and reader to the same convention. No multi-model Fortran-produced `LLt` fixture exists in the repo, so the multi-model writer layout rests on reading `amica15.f90:2308-2333` rather than on an oracle.
- `loadmodout`'s gm-based reordering does correctly permute `Lht` alongside `W`/`c`/`alpha` (`load.py:313-314`), so model indexing stays coherent. An earlier suspicion of a mismatch here was investigated and refuted.
- Every Fortran line citation in the new comments was checked against the cited ranges and is accurate, including correctly distinguishing `amica15.f90` (the reference binary's source) from `amica17.f90` (the later GG-only trim).

neuromechanist:
## Review findings addressed

All findings from the review above are fixed; none deferred. Five commits on top of the original:

| Commit | Fixes |
|---|---|
| `621c5db` | do_reject zero-sentinel parity + shared NumPy forward pass |
| `85f2a35` | torch LLt computed once after keep-best restore; warns when omitted |
| `07a12b6` | LLt documented in the wrapper's docstring |
| `750cdb8` | `write_amicaout` raises on a half-specified `Lht`/`Lt` pair |
| `c4358fe` | Seven new regression tests |

### The parity bug (critical)
Both backends now score only `good_idx` and scatter into zero-filled full-length arrays, so a rejected sample is exactly `0.0` in both `Lht` and `Lt`, matching Fortran's sentinel that `load_rej` depends on. Computing only the good set also means rejected outliers are no longer pushed through the model at write time.

### Shared forward pass, verified bit-exact
`_get_block_updates` and `_compute_full_posterior_ll` now share one `_forward_block` on the NumPy side (torch already shared `_forward`). This restores the `isfinite(logdet_W)` guard on the write path, which the duplicate had dropped.

Because this touches the training path in a bit-exact-parity codebase, it was verified against the pre-refactor commit rather than relying on a green suite: identical real-data fits were run on `1ab6797` and on the refactored tree, and the full LL trajectories compared with `np.array_equal`.

```
numpy     len  25 vs  25  array_equal=True  max|diff|=0.0
numpy_mm  len  15 vs  15  array_equal=True  max|diff|=0.0
torch     len  25 vs  25  array_equal=True  max|diff|=0.0
torch_mm  len  15 vs  15  array_equal=True  max|diff|=0.0
```

Bit-identical on all four, so the refactor is provably behavior-preserving and single-model #24 parity is untouched.

### Torch memory
`_X_t` is gone. `fit()` computes LLt once, strictly after the keep-best restore (so it reflects the parameters actually exported), and retains only the compact `(n_models+1, N)` arrays. The full sphered dataset is no longer pinned for the model's lifetime, which matters on CUDA.

### New tests (7)
`do_reject` zeroing on both backends (each asserts the fixture genuinely rejects something, so it cannot degrade into a tautology); NumPy multi-model; partial final block on both backends (the previous tests all used sample counts exactly divisible by `block_size`); `from_state_dict` -> `write_amica_output` asserting the warning fires and no LLt is written; and gm-reorder alignment, which pins `Lht`'s model axis to `W`/`gm` against a real non-identity gm ordering read from raw bytes. That last one closes a gap we found in our own test: `Lt == logsumexp(Lht, axis=0)` is permutation-invariant over the model axis and so could never have caught a misalignment.

Suite: **225 passed**, 9 skipped (MLX absent), 1 pre-existing xfail. Up 7 from 218, no regressions. `ruff`, `ruff format --check`, `ty` all clean.

### Deliberate divergences, documented not silent
- **LLt staleness.** Fortran fills `modloglik` during iteration i's E-step, the M-step then updates the parameters, and `write_output` writes both, so Fortran's on-disk LLt is stale by one M-step relative to its own written `W`/`A`. We recompute post-update, so ours is self-consistent with the written parameters. Immaterial near convergence and nothing downstream depends on the staleness, unlike the do_reject zero-sentinel, which is load-bearing and we do match. Commented at both call sites so nobody "fixes" it toward the wart.
- **Recompute cost.** Documented in the docstrings and tracked as #157 (stash per-sample `logV` during the E-step, as Fortran does), rather than left as a bare TODO.

Issue #136, opened 2026-07-14
Phase 4: MIR/PMI visualizations


neuromechanist:
Part of #133. Depends on Phase 1 and Phase 2.

matplotlib-based visualizations, a fresh design rather than a port of
postAmicaUtility's MATLAB/EEGLAB GUI plumbing:

- PMI heatmap (component-reordered via Phase 2's block-diagonal ordering),
  one panel per model.
- Scalp topography + generalized-Gaussian mixture PDF overlay per component
  (pyAMICA already exposes `alpha`/`mu`/`sbeta`/`rho`, a natural fit --
  equivalent in spirit to postAmicaUtility's `pop_topohistplot.m`).
- Multi-model posterior-probability timeline (equivalent in spirit to
  `modprobplot.m`).

Out of scope: general scrolling EEG viewers, ERP-conditioning plots, and other
non-AMICA-specific GUI utilities from postAmicaUtility.

neuromechanist:
Scope reduced during implementation: this phase ships two of the three planned plots. The scalp-topography plot (`plot_topo_pdf`) is deferred to **#159**.

**Shipped**, both verified against the MATLAB reference (PR #158):
- `plot_pmi_heatmap` — components-by-components pairwise MI, block-diagonally reordered. Our clean-room `pairwise_mi` matches the GPL `minfojp` at **r=0.9887** on identical signals.
- `plot_model_probability` — the `modprobplot` equivalent. Smoothed model probability matches `smooth_amica_prob` at **r=0.9886** (1 s) / 0.9594 (5 s); `P(model|data)` matches `LLt2v` to 1.4e-14; `v` is bit-exact vs `loadmodout15.m`. Visual gate accepted.

**Deferred to #159**: `plot_topo_pdf` needs source activations from a loaded `AmicaOutput`, which rests on an unsettled `W` convention — `loadmodout`'s `W` does not reproduce the live model's `transform()` sources at high fidelity under either orientation, even single-model with `c` identically zero (mean |corr| 0.877 / 0.932, 0/32 above 0.999). It is also the only one of the three with no MATLAB oracle, since `pop_topohistplot` is broken upstream. Rather than ship a plot that renders plausibly but cannot be verified, it was cut.

[...]

PR #158, opened 2026-07-17
Phase 4: MIR/PMI visualizations (PMI heatmap + model-probability timeline)

neuromechanist:
## Summary

Adds `pyAMICA/viz.py`: `plot_pmi_heatmap`, `plot_model_probability`, `plot_topo_pdf`. New top-level, backend-agnostic module consuming `AmicaOutput` from `loadmodout` (the path already verified bit-exact against `loadmodout15.m` in #155), so it works for any backend's output directory. Functions return a `Figure` and accept an optional `ax`/`axes`, unlike the legacy `numpy_impl/viz.py` which returns `None` and mutates pyplot global state; that module is left untouched.

Closes #136
Part of epic #133

## Licensing: why this was built the way it was

`pop_topohistplot.m` and `pop_modPMI.m` carry explicit **GPL-2.0-or-later** headers (Copyright Ozgur Baklan, SCCN, INC, UCSD); `modprobplot.m`, `minfojp.m`, `LLt2v.m`, `smooth_amica_prob.m` sit in the same GPL plugin. pyAMICA is BSD-3-Clause, which is why Phase 2's PMI was clean-room.

**No postAmicaUtility `.m` source was read at any point.** The behavioural spec came only from MATLAB `help` text (public API docs), rendered figures, and black-box input/output. Running the reference is what let us match observable behaviour without deriving from protected expression, so the MATLAB gate is not merely a check here, it is the mechanism that keeps the reimplementation clean. `pre_ICA_cleaning/getMIR.m` is Apache-2.0 and was legitimately read (Phase 1 ported it with attribution).

## MATLAB gate

Plots have no byte-level contract, so the gate is two-level: pin every plotted quantity to a MATLAB oracle, then compare rendered figures. Full record, traps and reproduction steps: `.context/issue-136/matlab_viz_verification.md`.

### Data gate: every plotted quantity is pinned

| Quantity | Oracle | Result |
|---|---|---|
| `mir()` | `getMIR.m` (Apache-2.0) | **1.7e-15** relative |
| P(model \| data) | `LLt2v` | **1.4e-14** vs our `softmax(Lht)` |
| `v` (log10 model odds) | `loadmodout15.m:284` | **bit-exact** |
| Hanning-smoothed probability | `smooth_amica_prob` | **r = 0.988577** (1 s), **0.959376** (5 s) |
| `pairwise_mi` | `minfojp` (GPL, black box) | **r = 0.9887** off-diagonal, identical signals |

The last two are correlations by design. Our PMI is a clean-room, genuinely different estimator, so equality would be suspicious; and near a model switch the probability is near-binary, so `max|diff|` ~ 1.0 is expected while mean abs diff is 0.014. **These must not be tuned toward MATLAB** — that would convert a clean-room reimplementation into a derivation.

This also closed a **Phase 1 gap**: MIR had only ever been validated against itself. The Apache-2.0 original was always available and is now run.

### Visual gate: accepted

`.context/issue-136/cmp_modprob.png` and `cmp_pmi.png` (committed). The timeline reproduces MATLAB's two panels, switching times, LL trace and seconds axis. The heatmap is rendered on **identical signals** (MATLAB's own `EEG.icaact`) so the comparison isolates estimator and ordering; both show the same dependent-subspace cluster and radiating cross pattern at the same MI scale. Accepted deliberate differences: viridis vs jet, we add a colorbar, 0-based labels, and a different (clean-room) ordering algorithm.

## Bug caught by the gate before it shipped

The obvious Hanning smooth, `np.convolve(row, w/w.sum(), mode="same")`, zero-pads beyond the data. Since `Lht ~= -108` is nowhere near 0, that padding drags the first and last half-window toward zero: input[0] = -123.4701 became **-60.39**. After the softmax that produces confidently **wrong** model probabilities at both ends of every plot: plausible-looking and completely wrong, the worst failure mode. Fixed by dividing by the window overlap (a shrinking normalised window at the boundaries). Guarded by a test asserting the edges are not dragged toward zero.

## Deliberate divergences (documented in-code so nobody "fixes" them)

- **Smoothing endpoints**: MATLAB pins the first/last sample to the raw *unsmoothed* value (a `smooth()` idiom). We return a locally averaged value. Ours is arguably more correct, and matching exactly would require reading GPL source.
- **MI diagonal**: MATLAB zeroes it; ours is self-entropy (~2.83 vs off-diagonal ~0.06), so `mask_diagonal=True` by default. Unmasked it destroys the colour scale.
- **Negative MI**: MATLAB's bias correction goes slightly negative; ours is strictly positive. Expected estimator difference.

## Test plan

Real bundled sample EEG and a real short 2-model fit throughout; no synthetic data (`.rules/testing.md`). `matplotlib.use("Agg")`. Tests assert on structure (axes count, labels, line count == n_models, probabilities summing to 1, tick labels equalling reordered component indices, masked diagonal, edge-not-dragged-to-zero, the `ValueError` and `ImportError` paths), not pixels and not merely "did not crash".

- [x] `pytest pyAMICA/tests/test_viz.py` -> **35 passed**
- [x] Full suite -> **260 passed**, 9 skipped (MLX absent), 1 pre-existing xfail. Up exactly 35 from the 225 baseline, no regressions.
- [x] `ruff check`, `ruff format --check`, `ty check` all clean
- [x] `mkdocs build --strict` passes with the new `docs/api/viz.md`


neuromechanist:
## Review summary

Four parallel reviewers plus a maintainer pass. The headline: the escalated "critical bug" was a **false positive caused by a wrong entry in my own verification doc**, but chasing it uncovered a **real, pre-existing, shipped bug** elsewhere.

### The false positive (and the doc that caused it)

A reviewer flagged `plot_topo_pdf`'s `out.W[:, :, model] @ sphered` as wrong, citing trap 3 of `.context/issue-136/matlab_viz_verification.md`, which claimed the "canonical" formula was `W[:, :, m].T @ (sphere @ (X-mean) - c[:, m])`. **That trap entry was wrong, I wrote it, and the reviewer escalated it in good faith.** Acting on it would have broken working code.

The rule, now corrected in the doc:

- `AmicaOutput.W` (from `loadmodout`) is the **EEGLAB convention: ROWS are components**, so `out.W @ sphere @ (X - mean)` is correct.
- `AMICATorchNG.W` (the live model) is the **internal convention: COLUMNS are components**, so its `transform` needs the transpose.

They are transposes of each other; neither formula ports to the other object. `load.py` proves the row convention: it normalises `A` by columns but `W` by **rows** (`W[i, :] *= na`), alongside `mu *= na` and `sbeta /= na` — mutually consistent only under the row reading, and exactly what `plot_topo_pdf` needs since it draws `A`'s column beside that component's activation and its `mu`/`sbeta` mixture.

The bad reasoning is recorded too, because it was seductive: best-match correlation against `transform()` ranked the wrong formula higher (0.988 vs 0.982 — both high, because `loadmodout` normalises and reorders), and the `std(s_i) = ||W[:, i]||` identity is true but proves nothing about intent. What settles it is consistency with `A`: `A @ (W @ sphered)` reconstructs the centred data to `0.000000` relative error, the transpose form to `0.894`.

### The real bug (`c502b44`) — pre-existing, shipped

`numpy_impl/pdf.py: compute_pdf` used `special.gammaln` where the generalized Gaussian needs `special.gamma`:

| rho | before | after |
|---|---|---|
| 1.0, 2.0 (special-cased) | 1.000000 | 1.000000 |
| **1.5 (the default `rho0`)** | **-8.82** | 1.000000 |

A density integrating to a negative number. Origin: the Fortran computes this in **log** space (`- gamln(1+1/rho) - log(2)`, `amica15.f90:1305-1306`) where log-gamma is correct; the port transcribed that `gamln` straight into a linear-space expression.

- **The fit path was never affected.** `numpy_impl/core.py` has its own log-space `_compute_log_pdf` mirroring the Fortran and never imports `.pdf`. Parity is untouched.
- The **pre-existing** `numpy_impl/viz.py: plot_pdf_fits` has been drawing wrong curves; this fixes it too.
- It survived because the only tests calling `compute_pdf` used **rho=1.0 and rho=2.0 — precisely the two special-cased branches that are correct**. The default 1.5 was never tested.

Now guarded by a property test (density integrates to 1 across rho 1.0–3.0, which no restatement of the formula could fake) and a continuity test (rho=2 vs rho=2+1e-9 must agree across the exact-equality branch boundary). **Both fail on the old code**; the pre-existing rho=1/rho=2 test passes on both, confirming it was structurally incapable of catching this.

### CI fix (`aa3017d`)

7 topo tests failed in CI because `mne` is the optional `viz` extra and CI runs bare `uv sync`. Fixed with a **per-test** skipif, not a module-level `importorskip`: the latter would also drop the 28 tests needing no mne, leaving a green suite that never exercised the module. Verified by simulating CI's environment: **28 passed, 7 skipped** (not 35 skipped).

### Also addressed (`3ab2544`)

- The smoothing-correlation comment now names *which* quantity it describes. An unlabelled `0.994` read as a stale copy of the doc's `0.9886` — they are different quantities (smoothed log-likelihood vs the resulting probabilities), both correct, and the ambiguity misled a reviewer.
- The `mask_diagonal` divergence note now carries the same "do not align this with MATLAB" imperative its smoothing counterpart has.
- Trap citation corrected (4 -> 5).

### Verified fine

`plot_model_probability` raises a clear `ValueError` when `out.Lht is None` — tested against a real LLt-stripped output directory (the #155 restored-model case), not by reading code.

Suite: **262 passed**, 9 skipped, 1 pre-existing xfail.

neuromechanist:
## Scope change: `plot_topo_pdf` cut, deferred to #159

Two plots ship (`plot_pmi_heatmap`, `plot_model_probability`), both fully pinned to MATLAB with the visual gate accepted. `plot_topo_pdf` is removed and tracked in **#159**.

### Why

It derives source activations from a loaded `AmicaOutput`, and that rests on a `W` convention that turns out to be unsettled. `loadmodout`'s `W` does not reproduce the live model's `transform()` sources at high fidelity under **either** orientation, even for a single-model fit where `c` is identically zero (verified `||out.c|| = 0.000000`):

| candidate | mean \|corr\| | n > 0.999 |
|---|---|---|
| `out.W @ sphered` | 0.877 | **0/32** |
| `out.W.T @ sphered` | 0.932 | **0/32** |

A merely normalised, variance-reordered `W` would give ~1.000 for all 32, because correlation already quotients out exactly the scale/sign/permutation `loadmodout` introduces. Neither does. `loadmodout` is separately bit-exact against EEGLAB's real `loadmodout15.m`, so this is not a simple bug; it is an unanswered question about what `out.W` *is*.

`plot_topo_pdf` is also the **only** plot with no MATLAB oracle (`pop_topohistplot` is broken upstream: `Unrecognized function or variable 'showhist'`), so nothing external could catch a wrong activation space. Shipping a scalp map that renders plausibly but cannot be verified is exactly the failure mode this epic keeps hitting, in a repo whose stated bar is numerical parity. Cut rather than documented-and-shipped.

The two plots that ship touch neither raw data nor sources, so they are unaffected.

### Three tests that look decisive and are not

Recorded in #159 and the gate doc, because each produced a confident wrong or useless answer during triage and the next person will reach for them:

1. **Activation mean vs fitted mixture mean** — degenerate. The fitted mixture is near-symmetric (`sum(alpha*mu) ~= 0.01`), so with `E[sphered]=0` the no-c formula matches trivially (0 ~= 0) regardless of correctness.
2. **Best-match correlation vs `transform()`** — cannot see `c` at all: Pearson correlation is shift-invariant, returning bit-identical numbers with and without it (0.849780 both).
3. **Histogram-vs-PDF L1** — too insensitive: it ranked no-c better (0.297 vs 0.326), but omitting `c` shifts each component by ~20% of its spread (`|wc|/std` mean 0.197, max 0.535), which should produce a far larger gap. Likely dominated by the unconverged fit. It was also silently invalid until the `gammaln` fix.

Review escalated this function twice as a critical bug, both times on the strength of a **wrong claim in our own verification doc** (now corrected). I disproved the specific claim, but could not positively verify the function, and did not want to merge on "I couldn't prove it wrong."

### The detour paid for itself

Chasing this is what surfaced the `gammaln`-for-`gamma` bug in `numpy_impl/pdf.py` (`c502b44`) — pre-existing, shipped, making the generalized-Gaussian density negative for every `rho` outside the special-cased 1 and 2, including the default `rho0=1.5`. That fix stays in this PR and repairs `numpy_impl/viz.py: plot_pdf_fits` too.

Suite: **258 passed**, 9 skipped, 1 pre-existing xfail.

Issue #159, opened 2026-07-17
loadmodout reads W, sbeta and rho in the wrong byte order (corrupts genuine Fortran output)


neuromechanist:
Surfaced while building Phase 4's `plot_topo_pdf` (#136), which needs per-component source activations derived from an `AmicaOutput`. Splitting it out rather than guessing, per the #155 precedent.

## The observation

`loadmodout`'s `W` does not reproduce the live model's own sources at high fidelity, under **either** orientation, even in the cleanest possible case.

Setup: real bundled sample EEG (32ch x 8192), `AMICATorchNG(n_models=1, n_mix=3, seed=1)`, 30 iterations. Single model, so `out.c` is identically zero (verified: `||out.c|| = 0.000000`), no gm model reordering, no component sharing. Ground truth is the model's own `ng.transform(data)`. Compared per-component by best `|corr|`, which is invariant to the per-component scale, sign and permutation that `loadmodout` legitimately introduces (it normalises to unit-norm `A` columns and variance-reorders):

| candidate | mean \|corr\| | min | n > 0.999 |
|---|---|---|---|
| `out.W[:, :, 0] @ sphered` | 0.877 | 0.47 | **0/32** |
| `out.W[:, :, 0].T @ sphered` | 0.932 | 0.81 | **0/32** |

where `sphered = out.S[:out.num_pcs] @ (data - out.data_mean[:, None])`.

If `out.W` were simply a normalised, variance-reordered `AMICATorchNG.W`, one of these should hit `|corr| ~ 1.000` for all 32 components, because correlation already quotients out exactly the transformations `loadmodout` applies. Neither does. 0.88/0.93 is "close but not it" (random would be ~0.1-0.2 in 32 dimensions), so this is not a permutation or scaling artefact.

Reproduced on a 2-model fit too (0.85 / 0.83 for model 0, 0.77 / 0.75 for model 1), with the gm model pairing explicitly checked and identity.

## Why this is not obviously a bug

`loadmodout` is verified bit-exact against EEGLAB's real `loadmodout15.m` on the genuine Fortran fixture and on pyAMICA-written output, single- and multi-model (`.context/issue-155/matlab_interop_verification.md`, `.context/issue-136/matlab_viz_verification.md`). Issue #92 separately verified the mixture-parameter layout through a MATLAB round-trip. So `out.W` is correct *in the EEGLAB sense*. The `A @ (out.W @ sphered)` reconstruction identity also holds to ~1e-15.

[...]

PR #160, opened 2026-07-17
Phase 5: wire MIR/PMI into the AMICA API

neuromechanist:
## Summary

Exposes the Phase 1-2 metrics on the model itself, per Scott Makeig's original request: computable at decomposition end, and optionally at specified decomposition waypoints.

- `AMICATorchNG.mir(X, *, model_idx=0, nbins=None) -> (mir_nats, variance)` and `.pmi(X, *, model_idx=0, nbins=None) -> mi_matrix`
- `AMICA.mir` / `AMICA.pmi` delegating through `_check_usable`, so they inherit the #50 degenerate-fit contract for free
- `fit(..., mir_step=0)` waypoint schedule populating `mir_history_` as `(iteration, mir_nats, variance)`

Closes #137
Part of epic #133

## Why this phase matters more than "convenience wiring"

Hand-composing the unmixing is **demonstrably** error-prone. While building Phase 4 I got it wrong, wrote the wrong rule into the verification doc, and a reviewer then escalated a non-bug as critical on that authority. #159 subsequently proved `loadmodout` itself reads `W`/`sbeta`/`rho` in the wrong byte order. A supported accessor removes a whole class of that.

The load-bearing test is `test_mir_composes_unmixing_the_documented_way`, which asserts `model.mir(X)` **exactly** equals `metrics.mir(get_unmixing_matrix(0) @ sphere, X)`. Verified it catches the real mistake: dropping the `.T` (the exact Phase 4 error) makes it fail. Note the numbers, because they explain why this was so easy to get wrong and so hard to spot — correct **42.5486** vs wrong **42.4688**, a 0.2% difference. A plausible number from a wrong matrix. An `allclose` test would sail past it; only exact equality against the documented composition catches it.

## Design notes

- **`mir` needs the full raw-data-to-sources transform**, `W[:, :, m].T @ sphere` (note the transpose), not bare `W`. MIR is shift-invariant, so the `mean`/`c` centering `transform` applies is irrelevant to it. `pmi` needs sources, so it calls `transform` rather than hand-rolling.
- **MIR-only waypoints**, on measured evidence: on the real 32x30504 sample, `mir` costs **20 ms** and `pairwise_mi` **678 ms** (34x), against a CPU fit iteration of ~36-400 ms. PMI on a schedule would dominate the fit, for a quantity that is a final structural diagnostic rather than a convergence signal. MIR over iterations is what actually answers the request. `model.pmi(X)` stays available on demand.
- **PCA guard**: `pcakeep`/`pcadb` leave the sphere rank-deficient, so MIR's `sum(log|eig|)` log-Jacobian is undefined. Both `mir()` and `mir_step>0` raise a clear `ValueError` naming the flags, up front rather than mid-fit, mirroring the existing `share_comps` guard (`torch_impl/core.py:553-561`) which rejects the same flags for the same underlying reason.
- **Waypoint gate follows the repo idiom**: `if mir_step > 0 and it % mir_step == 0`, the same shape as the NumPy backend's `writestep`/`histstep` (`numpy_impl/core.py:1429-1435`). `AMICATorchNG.fit` had no hook mechanism of any kind; no plugin system was invented.
- **`keep_best` (#51) interaction**: `mir_history_` is a true trajectory like `ll_history` and is *not* rewritten by a best-iterate rollback. The fit-end value is `model.mir(X)` (computed from the restored, exported parameters), **not** `mir_history_[-1]` — mirroring the existing `final_ll_` vs `ll_history_[-1]` distinction that exists precisely because of #51. Documented in-code.
- **Not persisted**: `mir_history_` stays out of `state_dict()` (a diagnostic, not a fitted parameter; adding it would need a format-version bump). `format_version` stays 3.
- **Not blocked by #159**: that bug is in the *loaded-fit* path (`loadmodout`). Everything here reads the live model's own `W`/`sphere` and `transform()`, which are correct. A loaded-fit accessor belongs to #159.

## Test plan

Real bundled sample EEG throughout, reusing the existing module-scoped `real_data`/`fitted_ng` fixtures; no synthetic data (`.rules/testing.md`).

- [x] `model.mir(X)` exactly equals the documented composition; `model.pmi(X)` exactly equals `pairwise_mi(model.transform(X))`
- [x] Real fitted unmixing gives `mir > 0` and `mir > mir(identity)`
- [x] `mir`/`pmi` added to the **existing** unfitted-raises and `test_degenerate_fit_refuses_output` lambda lists, rather than parallel new tests
- [x] `mir_step=2` populates `mir_history_` at [0, 2, 4] with finite values; `mir_step=0` leaves it empty and is byte-identical to omitting it; `mir_step<0` raises
- [x] `pcakeep` + `mir()` raises naming `pcakeep`; `pcakeep` + `mir_step>0` raises before fitting
- [x] Suite: **267 passed**, 9 skipped, 1 pre-existing xfail. Exactly +9 vs the 258 baseline, matching the 9 new tests.
- [x] `ruff check`, `ruff format --check`, `ty check` clean; `mkdocs build --strict` passes

No MATLAB gate this phase: MIR is already pinned to the Apache-2.0 `getMIR.m` oracle at 1.7e-15 and PMI to `minfojp` at r=0.9887 (Phase 4). This is wiring over already-verified metrics, and the tests assert the wiring reproduces the free functions exactly, so there is no new numerical claim to gate.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -5,6 +5,16 @@ Release notes are also published on the
 
 ## Unreleased
 
+- `AMICATorchNG`/`AMICA` gain `mir()`/`pmi()` accessors that compose the
+  fitted unmixing the documented way (`get_unmixing_matrix(model_idx) @
+  sphere` for MIR, `transform(X, model_idx)` for PMI) and delegate to
+  `pyAMICA.metrics.mir`/`pairwise_mi`, so callers no longer hand-compose the
+  transform themselves. `fit()` also accepts `mir_step` (default `0`, off) to
+  record MIR waypoints during training in `mir_history_` as
+  `(iteration, mir_nats, variance)`; like `ll_history_`, it is a true
+  trajectory that a `keep_best` restore does not rewrite. PCA reduction
+  (`pcakeep`/`pcadb`) is rejected up front with a named error, since it
+  leaves the sphere rank-deficient and MIR's log-Jacobian undefined (#137).
 - Visualization module (`pyAMICA.viz`): `plot_pmi_heatmap` and
   `plot_model_probability`, backend-agnostic views over `AmicaOutput` that return
   a `Figure` (and accept an optional `ax`/`axes`) rather than mutating pyplot
--- a/pyAMICA/amica.py
+++ b/pyAMICA/amica.py
@@ -70,7 +70,8 @@ class AMICA:
         ``ll_history_[-1]``, as the model's log-likelihood: with the best-iterate
         safeguard the returned parameters can be an earlier, higher-LL iterate.
     mir_history_ : list
-        MIR waypoint trajectory (issue #137), populated when ``fit`` is called
+        Mutual Information Reduction (MIR) waypoint trajectory (issue #137),
+        populated when ``fit`` is called
         with ``mir_step > 0``: ``(iteration, mir_nats, variance)`` tuples from
         the mid-fit ``W``/``sphere``. Like ``ll_history_``, a ``keep_best``
         restore does not rewrite it -- use :meth:`mir` on the fitted model for
@@ -357,7 +358,21 @@ class AMICA:
         Returns
         -------
         mir_nats : float
+            Mutual information removed, in nats.
         variance : float
+            Variance of the estimate.
+
+        Raises
+        ------
+        ValueError
+            If the model is unfitted; or if PCA reduction (``pcakeep``/
+            ``pcadb``) is active, which leaves the sphere rank-deficient so
+            MIR's log-Jacobian term is undefined; or if ``X`` is non-finite or
+            has a constant channel. See :meth:`AMICATorchNG.mir` and
+            :func:`pyAMICA.metrics.mir`.
+        RuntimeError
+            If the fit ended degenerate (issue #50), since the parameters are
+            non-finite and any metric from them would be meaningless.
         """
         self._check_usable("compute MIR")
         assert self.model_ is not None
@@ -385,6 +400,19 @@ class AMICA:
         Returns
         -------
         mi_matrix : np.ndarray of shape (n_sources, n_sources)
+            Symmetric pairwise mutual information, in nats. The diagonal is each
+            source's own entropy, not a mutual information; see
+            :func:`pyAMICA.viz.plot_pmi_heatmap`, which masks it by default.
+
+        Raises
+        ------
+        ValueError
+            If the model is unfitted; or if ``X`` is non-finite or has a
+            constant channel; or if ``nbins`` is too large for the sample count
+            (a sparse joint histogram fabricates mutual information from noise).
+            See :func:`pyAMICA.metrics.pairwise_mi`.
+        RuntimeError
+            If the fit ended degenerate (issue #50).
         """
         self._check_usable("compute PMI")
         assert self.model_ is not None
--- a/pyAMICA/amica.py
+++ b/pyAMICA/amica.py
@@ -69,6 +69,12 @@ class AMICA:
         Log-likelihood of the *fitted* parameters (issue #51). Use this, not
         ``ll_history_[-1]``, as the model's log-likelihood: with the best-iterate
         safeguard the returned parameters can be an earlier, higher-LL iterate.
+    mir_history_ : list
+        MIR waypoint trajectory (issue #137), populated when ``fit`` is called
+        with ``mir_step > 0``: ``(iteration, mir_nats, variance)`` tuples from
+        the mid-fit ``W``/``sphere``. Like ``ll_history_``, a ``keep_best``
+        restore does not rewrite it -- use :meth:`mir` on the fitted model for
+        the value of the *returned* parameters, not ``mir_history_[-1]``.
 
     Examples
     --------
@@ -107,6 +113,7 @@ class AMICA:
         self.final_ll_ = None
         self.stop_reason_ = None
         self.converged_ = False
+        self.mir_history_ = []
 
     def _select_device(self, ng_dtype) -> Union[str, torch.device]:
         """Resolve the compute device, applying the MPS/float64 fallback.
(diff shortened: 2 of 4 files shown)


neuromechanist:
## Review summary

Four reviewers, each in its **own isolated worktree** this time (sharing one worktree across concurrent reviewers in #158 manufactured three phantom findings, so that is fixed). One real bug, one flaw in my own test design, and two doc nits.

### The real bug (`9343656`): a failed waypoint destroyed the fit

The waypoint called `self.mir(X)` unprotected, so a raising waypoint propagated straight out of `fit()` and threw away the whole decomposition. Verified by forcing the raise: the `ValueError` escaped `fit()` and the run was lost.

This was **reachable, not theoretical**. `metrics.mir` raises on a near-singular unmixing, and a near-singular W mid-fit is a transient the natural gradient can pass through — the training path only *warns* about the same condition (`numpy_impl/core.py:827`). So enabling a purely diagnostic flag could throw away hours of training over something the optimiser was about to recover from.

Now caught at the waypoint: warn naming the iteration and cause, and record `(iter, nan, nan)` so the gap stays **visible** in `mir_history_` rather than silently absent — a plotted trajectory shows a hole exactly where the transient was. Not a silent failure. `test_failing_mir_waypoint_does_not_kill_the_fit` fails on the unprotected version.

### CI caught a flaw in my own test design

I had insisted on exact equality for the convention test. It passed locally and **failed on CI**:

```
assert (42.548551297...9056413769711) == (42.548551297...9056413769711)
```

Identical to ~12 digits, differing in the last bits: `mir()` composes `W.T @ sphere` in torch while the test composes it in numpy, and the two BLAS paths disagree at the ULP level. Relaxed to `rtol=1e-12`, which is still **nine orders of magnitude** tighter than the 1.9e-3 error it guards (dropping the `.T` gives 42.4688 vs 42.5486). Verified it still fails on a dropped `.T`. The guard is intact and now portable.

### Checked and cleared (verified, not assumed)

- **Degenerate fit**: refused at *both* layers — `RuntimeError: Refusing to compute MIR: fit ended degenerate` at the wrapper, `LinAlgError: Array must not contain infs or NaNs` at the backend. No plausible-but-wrong number, which was the main risk for a metric API.
- **NaN/Inf in X**: clear `ValueError` from `validate_signal_matrix` for both `mir` and `pmi`, inherited from the Phase 1/2 metric hardening.
- **Docs**: the comment reviewer found no inaccurate claims and specifically confirmed this diff does **not** repeat the epic's W-convention mistake, asserts nothing about `AmicaOutput`/`loadmodout` (leaving #159 untouched), and states the keep_best caveat in three places a user would see.

### Doc fixes

- "MIR" now expanded at its true first use in each file (AGENTS.md: define abbreviations before use). It appeared unexpanded in the `AMICA` class docstring, which is the first thing `help(AMICA)` shows.
- `AMICA.mir`/`AMICA.pmi` gained `Raises` sections: a user reading only the wrapper had no way to learn about the `pcakeep`/`pcadb` `ValueError`, which lives on the backend method.

Suite: **268 passed**, 9 skipped, 1 pre-existing xfail.

neuromechanist:
## Review round 2: remaining findings resolved (`9b9c9bd`)

### Two real test gaps, proven by sabotage

`model_idx` and `nbins` were **never exercised**. Hardcoding either one inside `mir()` and `pmi()` left all nine tests passing, because the shared fixture is single-model and every wiring test defaulted to model 0 / `nbins=None` on *both* sides of the comparison. Fixed with a genuine 2-model fit and a non-default `nbins`; verified both new tests fail on that exact sabotage:

```
model_idx hardcoded to 0  -> FAILED test_mir_and_pmi_honour_model_idx: model_idx=1 returned model 0's MIR
nbins hardcoded to None   -> FAILED test_mir_and_pmi_honour_nbins: nbins was ignored by pmi()
```

### A test that could not test its own claim

`test_mir_step_raises_under_pca_reduction_up_front` matched on `"pcakeep"` — which **both** guards mention, `fit()`'s up-front one and `mir()`'s downstream one. So it could not distinguish "rejected before iteration 0" from "rejected during iteration 0", which is the entire claim it exists to make. Now matches `"Rejected up front"`, unique to `fit()`'s guard.

(The waypoint fix in `9343656` incidentally hardened this: with the ValueError now caught at the waypoint, deleting the up-front guard yields no error at all rather than a look-alike message. The tighter match keeps the test honest independently of that.)

### One CRITICAL refuted, with evidence

A reviewer flagged as critical that `pmi()` returns a finite matrix on a singular W while `mir()` refuses — the "plausible but wrong" failure mode. **Tested, and it is a false positive.** With W made exactly singular (min |eig| = 4.8e-17):

```
sources 3 and 5 are now identical:   max|s3 - s5| = 0.00e+00
MI(3,5) = 2.7478   <- the collinear pair, and the MAXIMUM off-diagonal
H(s3)   = 2.7478   <- MI(3,5)/H(s3) = 1.0000 exactly
median off-diagonal MI = 0.0291     (MI(3,5) is 94x this)
```

That is not a plausible-but-wrong number: it is the mathematically correct answer, `MI(X,X) = H(X)`, the theoretical maximum. It *is* the degeneracy alarm. The asymmetry with `mir()` is principled rather than a gap: MIR needs an invertible transform (its `log|det W|` term is undefined for a singular map, so refusing is correct), while PMI is mutual information between signals and stays well-defined for degenerate sources. Adding the proposed conditioning guard to `pmi()` would reject a correct and diagnostically valuable answer.

### Deferred (not blocking)

`mir_step` x `keep_best` rollback, and `mir_history_` after save/load. Both are documented claims without direct tests. They are diagnostic-trajectory bookkeeping on an opt-in feature, and the underlying mechanism (`mir_history_` is absent from `_PARAM_TENSORS`, so `_restore_params` cannot touch it) was verified by review. Worth a follow-up rather than holding the epic's final phase.

Suite: **270 passed**, 9 skipped, 1 pre-existing xfail.

Issue #161, opened 2026-07-17
Test mir_history_ against the keep_best rollback and save/load


neuromechanist:
Follow-up from #137 / PR #160. Two documented claims about `mir_history_` have no direct test. Neither is believed broken; both are verified by review only, and this epic has repeatedly found that a documented-but-untested claim is exactly where a regression hides.

## 1. `mir_history_` vs the `keep_best` rollback (#51)

The docstrings state plainly (`torch_impl/core.py`, `amica.py`) that `mir_history_` is a **true trajectory** which a best-iterate restore does NOT rewrite, and therefore that the fit-end MIR is `model.mir(X)` (computed from the restored, exported parameters), **not** `mir_history_[-1]`. This mirrors the existing `final_ll_` vs `ll_history_[-1]` distinction that exists precisely because of #51.

Verified by review: `mir_history_` is absent from `_PARAM_TENSORS`, and `_snapshot_params`/`_restore_params` only clone and restore that tuple, so the rollback cannot touch it. But nothing exercises a fit where a restore actually fires.

This is a **known-reachable** gap, not hypothetical: `test_write_amica_output_ll_matches_kept_iterate` in the same file already has a proven recipe for forcing a real rollback on real data (`n_models=2, max_iter=60, do_newton=True, newt_start=1, lrate=0.5, seed=0`, with a `pytest.skip` fallback if that seed happens to stay monotone). Extend it to `mir_step > 0` and assert the trajectory survives the restore intact.

## 2. `mir_history_` after save/load

`AMICA.load()` comments that `mir_history_` is deliberately not persisted in `state_dict()` (it is a diagnostic, not a fitted parameter; persisting it would need a format-version bump) and is therefore always empty on a loaded model. Nothing does `fit(mir_step>0)` -> `save` -> `load` -> assert empty. A regression that leaked stale trajectory data through `state_dict`/`from_state_dict` would go unnoticed.

## Also worth a docstring line (informational, from review)

Within the fit loop, `mir_history_`'s entry for iteration `i` is computed from the parameters **after** that iteration's update, while `ll_history[i]` records the LL of the **pre-update** parameters at the same `i` (see its own comment). So the two histories describe snapshots one update apart, and a reader zipping them by index would be comparing different parameter states. Intentional and consistent with the code, but nowhere stated.

## Why deferred rather than blocking

[...]

Issue #157, opened 2026-07-16
Stash per-sample logV during the E-step instead of recomputing for LLt


neuromechanist:
Follow-up from #155 / PR #156, which added Fortran-parity `LLt` output (per-timepoint, per-model log-likelihood) to both backends.

## Current design

Neither backend retains a full-dataset per-sample log-likelihood after the E-step: it is computed per block inside `_get_block_updates` and discarded once folded into sufficient statistics. So #156 recomputes it with a fresh full-dataset forward pass at write time. This was the right call for that PR (it is correct after a `keep_best` rollback, and it avoids retaining the sphered dataset), but it has two known costs.

## Costs

1. **Repeated recompute on the NumPy backend.** `_write_results` runs at every `writestep` checkpoint during training, not only at the end (default `writestep=100`; the bundled sample config uses 20). Each checkpoint now pays a full O(n_samples) forward pass it did not pay before. Bounded and small at the default, and NumPy is the legacy backend, which is why #156 accepted it.
2. **One extra full E-step per torch fit.** `fit()` computes LLt once at the end even if `write_amica_output` is never called.

## Cheaper design

Stash the per-sample `logV` as it is computed during the training E-step and reuse it at write time, which is what Fortran does: it keeps `modloglik(num_models, N)` permanently allocated (`amica15.f90:2601`) and `write_output` just dumps it. Memory cost is `(n_models+1) * N * 8` bytes, far smaller than the sphered data.

Two attractive side effects:
- It reproduces the `do_reject` zero sentinel for free. Fortran's E-step only writes `modloglik` at good indices, so rejected samples naturally retain their zero-initialized value; #156 has to arrange that explicitly by scattering into zero-filled arrays.
- It matches Fortran's write cadence exactly.

## Why it was not done in #156

It collides with the torch backend's `keep_best` rollback (#51): a stashed `logV` from the last training iteration would not correspond to the restored best iterate's parameters. Making it work means snapshotting the stashed `logV` alongside the parameters in `_snapshot_params`/`_restore_params`, which is a real design change to the best-iterate mechanism and was out of scope.

[...]

PR #154, opened 2026-07-16
Phase 2: PMI core metric and component reordering

neuromechanist:
## Summary

Adds a clean-room implementation of Pairwise Mutual Information (PMI) between AMICA components, plus a greedy block-diagonal reordering, as `pyAMICA.metrics.pairwise_mi(sources, nbins=None) -> np.ndarray` and `pyAMICA.metrics.block_diagonal_order(mi_matrix) -> np.ndarray`.

This is deliberately NOT a port. The historical MATLAB reference (`minfojp.m`/`get_mi.m`/`arrminf2.m` in `sccn/postAmicaUtility`) is GPL-2.0-or-later, incompatible with a direct port into this BSD-3-Clause project. Per the epic's licensing decision, this implementation was derived solely from the algorithm description in issue #135 (binned marginal/joint entropy with a Miller-Madow bias correction; a greedy block-diagonal reordering) and standard statistics/algorithms -- the GPL source was never fetched or read at any point in this implementation.

`pairwise_mi` computes each pair's joint 2D histogram and derives both marginals by summing that same histogram (rather than independently rebinning), keeping the entropy terms self-consistent; entropy uses a proper Miller-Madow correction (`(m_occupied - 1)/(2N)`), distinct from MIR's fixed `(nbins-1)/(2N)` term. `block_diagonal_order` is a standard greedy nearest-neighbor chain seriation, matching the issue's own "greedy" framing.

Also includes a small phase-1 follow-up: extracted the shared input-validation logic (non-finite check, constant-channel check, nbins resolution) from `mir.py` into a new private `pyAMICA/metrics/_common.py`, reused by both modules -- a pure refactor, the 11 existing MIR tests pass unchanged. Also added the `delorme2012independent` citation (already in this repo's `paper.bib`) to `mir.py`'s docstring, which the epic's own background section ties MIR to but the phase 1 docstring omitted.

Closes #135
Part of epic #133

## Test plan

- `pyAMICA/tests/test_metrics_pmi.py` (new, 8 tests): symmetry, diagonal-matches-self-entropy, the real scientific claim that ICA sources have lower pairwise MI than raw channels (real fitted `pyresults/W.npy`/`sphere.npy` vs raw real channels), valid-permutation and adjacent-weight-improvement checks for the reordering, and the same non-finite/constant-channel error-path tests as phase 1.
[...]

Issue #34, opened 2026-07-03
Epic: Clean module naming and structure (drop amica_/pyAMICA/_torch/_v2 sprawl)


neuromechanist:
## Motivation
The module layout grew organically and is hard to navigate: NumPy code is split across
`pyAMICA.py` + `amica_*.py` prefixed modules, and the PyTorch code carries `amica_torch.py`,
`amica_torch_ng.py`, `amica_torch_v2.py`. We want clean, topic-based names (`newton`, `pdf`,
`data`, ...) shared conceptually between backends, with PyTorch counterparts where needed, and no
redundant `amica_`/`_torch`/`_v2` prefixes. Clean code is a first-class goal, not cosmetic.

This is intentionally **separate from the parity work** (epic #9 / PR #33) so renames don't muddy
the parity diff.

## Proposed target structure
```
pyAMICA/
├── amica.py            # public scikit-learn-style interface (unchanged import path)
├── version.py
├── numpy_impl/         # (or keep flat) legacy NumPy reference
│   ├── core.py         # was pyAMICA.py (AMICA_NumPy)
│   ├── newton.py       # was amica_newton.py
│   ├── pdf.py          # was amica_pdf.py
│   ├── data.py         # was amica_data.py
│   ├── load.py         # was amica_load.py
│   ├── viz.py          # was amica_viz.py
│   ├── utils.py        # was amica_utils.py
│   └── cli.py          # was amica_cli.py
└── torch_impl/
    ├── core.py         # was amica_torch_ng.py (AMICATorchNG, the parity backend)
    ├── newton.py       # was newton_optimizer.py
    ├── pdf.py          # was adaptive_pdf.py
    ├── mixture.py      # was mixture_models.py
    ├── optim.py        # was optimizers.py
    ├── output.py       # was fortran_output.py
    └── utils.py        # unchanged
```
(Exact package split -- flat vs `numpy_impl/` -- to be decided in phase 1.)

## Constraints
- Keep the public import surface stable: `from pyAMICA import AMICA, AMICA_NumPy, AMICATorchNG`.
- No backward-compat shims / re-export stubs for old module paths (per project rules); update all
  call sites, tests, CLI, and docs directly.
- Real-data tests stay green throughout; parity unchanged.

## Suggested phasing (sub-issues to be created when this starts)
1. Decide package layout + rename the PyTorch `torch_impl/` modules; update imports/tests/docs.
[...]

PR #162, opened 2026-07-17
Epic: MIR and PMI separation-quality metrics

### LINKED ISSUE #133: Epic: port MIR and PMI separation-quality metrics from postAmicaUtility/pre_ICA_cleaning

neuromechanist:
## Summary

Ports the MIR and PMI separation-quality metrics into pyAMICA, wires them into the model API, and adds two MATLAB-verified visualizations. Five phases, each individually reviewed and merged into this branch.

Closes #133

| Phase | PR | What |
|---|---|---|
| 1 | #153 | `mir` (mutual information reduction), ported from Apache-2.0 `getMIR.m` |
| 2 | #154 | `pairwise_mi` + `block_diagonal_order`, clean-room from a GPL reference |
| 3 | #156 | `LLt` posterior output parity with Fortran (both backends) |
| 4 | #158 | `plot_pmi_heatmap`, `plot_model_probability` |
| 5 | #160 | `model.mir(X)` / `model.pmi(X)` + `mir_step` waypoints |

## Licensing

pyAMICA is BSD-3-Clause, and the two references have incompatible licences, so they were handled differently and deliberately:

- **MIR** is a permitted direct port of `getMIR.m` (**Apache-2.0**, bigdelys/pre_ICA_cleaning), pinned to a commit SHA, with the full licence vendored in `THIRD_PARTY_NOTICES.md` to satisfy Apache-2.0 section 4(a).
- **PMI** is a **clean-room reimplementation**. Its reference (`minfojp.m`/`get_mi.m`/`arrminf2.m` in postAmicaUtility) is **GPL-2.0-or-later**; that source was never fetched or read, and a reviewer independently checked for leakage.
- **Phase 4's visualizations** are a fresh design. postAmicaUtility's `pop_topohistplot.m`/`pop_modPMI.m` carry explicit GPL headers, so they were **run and observed, never read**: `help` text, rendered figures and black-box I/O only. That posture is what let us match observable behaviour without deriving from protected expression.

## Verification against MATLAB

Every quantity that could be pinned to an external oracle was pinned:

| Quantity | Oracle | Result |
|---|---|---|
| `mir()` | `getMIR.m` (Apache-2.0) | **1.7e-15** relative |
| `LLt` reader/writer | real `loadmodout15.m`, MATLAB R2025b | **bit-exact**, both directions |
| `P(model \| data)` | `LLt2v` | **1.4e-14** |
| `v` (log10 model odds) | `loadmodout15.m` | **bit-exact** |
| Smoothed model probability | `smooth_amica_prob` | **r = 0.9886** |
| `pairwise_mi` | `minfojp` (GPL, black box) | **r = 0.9887** |

The MIR oracle closed a Phase 1 gap: MIR had only ever been validated against itself, while the Apache-2.0 original was available the whole time.

## Bugs found and fixed along the way

- **`do_reject` LLt parity** (#156): both backends wrote real log-likelihoods where Fortran writes exactly `0.0`. Those zeros are load-bearing, not padding: Fortran's `load_rej` reconstructs the rejection mask from them, so a Fortran run reading a pyAMICA `LLt` would have concluded nothing was rejected.
- **`gammaln` for `gamma` in `numpy_impl/pdf.py`** (#158): pre-existing and shipping. Made the generalized-Gaussian density **negative** for every rho outside the special-cased 1 and 2 — including the default `rho0=1.5`. It survived because the only tests called `compute_pdf` with rho=1.0 and 2.0, exactly the two hard-coded-correct branches. The fit path was never affected (it has its own correct log-space implementation), but the shipped `plot_pdf_fits` was drawing wrong curves.
- **`loadmodout` reshape order** (#156): read the `LLt` bytes C-order where the layout is column-major.
- **A failed MIR waypoint destroyed the fit** (#160): caught pre-merge; a diagnostic now warns and records a visible NaN gap instead of aborting a recoverable run.

## Deliberate divergences, documented not silent

- **LLt staleness**: Fortran's on-disk `LLt` is stale by one M-step relative to the `W`/`A` beside it. We recompute post-update, so ours is self-consistent. That wart is not load-bearing, unlike the `do_reject` zero-sentinel, which we do match.
- **Smoothing endpoints**: MATLAB pins the first/last sample to the raw unsmoothed value; we return a locally averaged one.
- **MI diagonal**: MATLAB zeroes it; ours is self-entropy, so the heatmap masks it rather than fabricating a zero.

## Scope changes

- **`plot_topo_pdf` was cut** to #159. It needs source activations from a loaded `AmicaOutput`, which turned out to rest on a real loader bug (below). It was also the only plot with no MATLAB oracle, so nothing external could catch a wrong activation space.
- **Follow-ups filed**: #157 (cheaper LLt recompute), **#159** (`loadmodout` reads `W`/`sbeta`/`rho` in the wrong byte order — corrupts genuine Fortran output; found while chasing #158, proven by recomputing the Fortran's own reported LL), #161 (two untested `mir_history_` claims).

## Test plan

Real bundled sample EEG and the real Fortran fixture throughout; no synthetic data as the basis of any correctness claim (`.rules/testing.md`).

- [x] **270 passed**, 9 skipped (MLX/mne absent), 1 pre-existing xfail on the integrated branch
- [x] Fortran-parity suite (`-m slow`) run on the integrated branch
- [x] `ruff`, `ty`, `mkdocs build --strict` clean
- [x] Phase 3's shared-forward-pass refactor touches the training path, so it was proven **bit-identical** against the pre-refactor commit (`max|diff| = 0.0` across four real-data LL trajectories) rather than trusted to a green suite


neuromechanist:
## Integration review

Three reviewers over the **cumulative** `main...HEAD` diff, each in its own isolated worktree, briefed to hunt for what the *integration* broke rather than to re-review the phases. That framing paid off: every finding below is invisible to a per-phase review, because each phase PR was only ever read against its own base.

### Fortran parity: PASSES on the integrated branch

The check that actually matters for this repo, and it does not run in CI:

```
uv run pytest pyAMICA/tests/ -m "slow"
4 passed, 1 skipped, 2 xfailed in 1887.39s (0:31:27)
pyAMICA/tests/torch_tests/test_ng_backend.py ...  [100%]
```

All three end-to-end Fortran-parity tests pass, including `test_full_fit_parity_numpy_vs_ng`. This matters most because Phase 3 refactored the **training path** (extracting a shared `_forward_block`). Both xfails are pre-existing: `git diff main...HEAD -- pyAMICA/tests/test_sample_data.py` shows this epic made no xfail changes.

### Fixed: three claims the `plot_topo_pdf` cut invalidated (`0586ae0`)

1. **`viz.py`'s module docstring still announced "three plots"** and described the topography view, while the module defines two. The commit that cut the plot missed the docstring, so a confidently wrong claim survived the exact commit that invalidated it. That is the third time this epic has been bitten by that pattern.
2. **`read_eeglab_set_metadata` refused any file with an unlocalized channel.** That guard was correct when its only consumer was the scalp-topography plot, which genuinely could not use one. That plot is gone, so the guard blocked the sample rate (its one remaining purpose) over a field nothing reads. EOG channels are commonly unlocalized, so this would have hit real datasets. Now NaN with a documented contract, guarded by a test that blanks one channel in a copy of the real bundled `.set` and fails if the old behaviour returns.

### Fixed: the changelog was missing three of five phases (`b5323da`)

Verified commit-by-commit: **Phases 1, 2 and 3 added no changelog entry at all.** Only Phases 4 and 5 did. A user reading it would never learn that MIR, PMI, or the `LLt` output shipped, the last of which is a new on-disk file with a MATLAB-verified EEGLAB round trip. Nobody noticed because nobody had read the five phases together. Also:

- `docs/guides/eeglab.md`'s output-file table, the page someone integrating with EEGLAB actually reads, still listed the pre-`LLt` file set.
- `pyAMICA.metrics` had **no API page**, violating `.rules/documentation.md`'s own one-page-per-module convention, while `docs/api/viz.md` already linked to `pyAMICA.metrics.pairwise_mi` with nothing to resolve to. It is the computational core Phases 1-2 shipped.

### Verified clean

- **Collection is complete**: 271 passed / 9 skipped / 1 pre-existing xfail, with per-file counts confirmed. This epic was bitten twice by tests silently not running (a stale base hiding 28 metrics tests; an optional dependency), so this was checked explicitly rather than assumed.
- **Zero deletions in any test file** across the whole epic (`--numstat`): no assertion was weakened by a later phase.
- **The `plot_topo_pdf` cut is otherwise complete**: no dangling imports, exports, nav entries, orphaned tests, or dead helpers.
- **The three previously-fixed structural-gap tests are still fixed** in the integrated tree.
- **No code in this epic depends on #159's broken loader path**: `plot_pmi_heatmap` takes a raw matrix, `plot_model_probability` touches only `out.Lht`, and `mir()`/`pmi()` read the live in-process model.
- `__all__` matches actual exports; `pyproject.toml` picked up `pyAMICA.metrics` (avoiding the #34 wheel-drop bug); `mkdocs build --strict` clean.

PR #163, opened 2026-07-17
Fix loadmodout byte order for W/sbeta/rho (#159)

### LINKED ISSUE #159: loadmodout reads W, sbeta and rho in the wrong byte order (corrupts genuine Fortran output)

neuromechanist:
## Summary

`loadmodout` read the Fortran/EEGLAB interop files `W`, `sbeta` and `rho` in C order while the writer, genuine Fortran output, and EEGLAB's `loadmodout15.m` all use column-major (F order). The result: `AmicaOutput.W` came back transposed, silently corrupting genuine Fortran output and everything derived from it (`A = pinv(W@S)`, `svar`, `origord`), and `sbeta`/`rho` were scrambled whenever `num_mix > 1` (the default). A write-then-read round trip cancels the error, so no self-consistency test could catch it, which is why prior gates (#92, #155) missed it.

Closes #159.

## What changed

- **Reader** (`loadmodout`): `W`, `sbeta`, `rho` now read `order="F"`, matching the writer, genuine Fortran, and `loadmodout15.m`. This also corrects the derived `A`/`svar`/`origord`.
- **Writer** (`write_amicaout`): the 3D `W` is written model-axis-slowest (`transpose(2, 0, 1)`) so multi-model output is genuine Fortran and EEGLAB-readable. **Single-model output is byte-identical to before** (verified), preserving #92.
- **`load_results`**: reads the same genuine-Fortran bytes and transposes back to the internal-backend convention it returns for the NumPy viz helpers (identical output for single-model; fixes latent multi-model breakage).
- **`AmicaOutput.sources(X, model=0)`**: new supported accessor (loaded-fit counterpart of the live model's `transform`), so downstream source derivation no longer hand-rolls the sphere/unmixing composition, the error class that spawned this issue.
- **`compare_with_fortran`** (uncalled diagnostic): transposes one side so it stays apples-to-apples.

## Why it is correct (two external oracles)

1. **LL recomputation**: recomputing the bundled fixture's own reported converged LL (-3.4018730) from what `loadmodout` reads gives -3.4018468 (5 sig figs); the shipped C-order read gave -3.5167 (off by 0.11). LL is not transpose-symmetric, so unlike `A@W==I` it discriminates orientation.
2. **MATLAB gate**: EEGLAB's real `loadmodout15.m` vs Python `loadmodout` for `W`/`A`/`sbeta`/`rho`, on the genuine single-model fixture **and** a pyAMICA 2-model output, agree to <=1.3e-13 (`.context/issue-159/matlab_interop_verification.md`). This is the only test that pins direction 2 of the interop contract.

## Test plan

- New regression tests (all sabotage-validated to fail on the buggy read):
  - `test_loadmodout_reproduces_fortran_reported_ll` (LL oracle, real fixture)
  - `test_loadmodout_sources_reproduce_live_transform_multimodel` (2-model round trip: `sources` matches live `transform` at |corr| > 0.999 per model; was 0/32 before)
  - `test_amicaoutput_sources_contract_and_c_subtraction`
  - `test_load_results_returns_internal_w_multimodel`
- Full non-slow suite: 275 passed, 9 skipped, 1 xfailed, no collection drop.
- `test_sample_data_cli` (strict xfail, 2000-iter) still xfails (the W flip did not turn it into an XPASS).
- ruff, ty, `mkdocs build --strict` clean.

## Follow-up

`plot_topo_pdf` (deferred out of #136) is now unblocked and can build on `AmicaOutput.sources`.

### DIFF
--- a/.context/issue-159/gate_check.py
+++ b/.context/issue-159/gate_check.py
@@ -1,0 +1,36 @@
+"""MATLAB gate check for issue #159 (run last).
+
+Compares Python ``loadmodout`` against MATLAB ``loadmodout15.m`` element-wise for
+W/A/sbeta/rho, on the genuine single-model fixture and the pyAMICA 2-model output.
+Both are the same algorithm (Python is a port), so a correct byte order makes them
+agree to floating-point noise; the pre-#159 C-order read made W disagree by the
+internal transpose (and A/svar with it), and sbeta/rho disagree for num_mix > 1.
+"""
+
+import os
+import numpy as np
+from scipy.io import loadmat
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+TOL = 1e-9
+
+ok = True
+for tag in ("fixture", "two_model"):
+    py = np.load(os.path.join(HERE, f"py_{tag}.npz"))
+    mat = loadmat(os.path.join(HERE, f"mat_{tag}.mat"))
+    print(f"\n== {tag} (num_models={int(py['num_models'])}) ==")
+    for name in ("W", "A", "sbeta", "rho"):
+        p = np.asarray(py[name])
+        m = np.asarray(mat[name])
+        # MATLAB drops trailing singleton dims (num_models==1 -> 2-D); align.
+        if m.ndim < p.ndim:
+            m = m.reshape(p.shape)
+        close = p.shape == m.shape and np.allclose(p, m, atol=TOL, rtol=1e-7)
+        diff = np.max(np.abs(p - m)) if p.shape == m.shape else float("nan")
+        ok &= close
+        print(
+            f"  {name:6s} py{p.shape} mat{m.shape}  max|diff|={diff:.2e}  "
+            f"{'OK' if close else 'MISMATCH'}"
+        )
+
+print("\nGATE", "PASS" if ok else "FAIL")
--- a/.context/issue-159/gate_compare.m
+++ b/.context/issue-159/gate_compare.m
@@ -1,0 +1,29 @@
+% MATLAB gate for issue #159 (run after gate_prep.py, before gate_check.py).
+% Loads the genuine single-model Fortran fixture and the pyAMICA-written 2-model
+% output with EEGLAB's real loadmodout15.m, and saves W/A/sbeta/rho so the Python
+% companion can check they match what Python loadmodout read. This is the only
+% test that pins direction (2) of the interop contract: EEGLAB reads pyAMICA's
+% (and Fortran's) bytes correctly. A pyAMICA write->read round trip cannot.
+
+here = fileparts(mfilename('fullpath'));
+root = fullfile(here, '..', '..');
+sample = fullfile(root, 'pyAMICA', 'sample_data');
+
+% addpath sample_data LAST so its working loadmodout15.m shadows any broken
+% copy elsewhere on the path (postAmicaUtility's has a syntax error on R2025b).
+addpath(sample);
+
+dirs = struct('tag', {'fixture', 'two_model'}, ...
+              'path', {fullfile(sample, 'amicaout'), fullfile(here, 'gate_2model')});
+
+for k = 1:numel(dirs)
+    % loadmodout15 concatenates paths directly, so the dir needs a TRAILING SLASH.
+    d = [dirs(k).path filesep];
+    m = loadmodout15(d);
+    W = m.W; A = m.A; sbeta = m.sbeta; rho = m.rho; num_models = m.num_models; %#ok<NASGU>
+    save(fullfile(here, ['mat_' dirs(k).tag '.mat']), 'W', 'A', 'sbeta', 'rho', 'num_models');
+    fprintf('%s: num_models=%d  W=%s  A=%s\n', dirs(k).tag, m.num_models, ...
+            mat2str(size(m.W)), mat2str(size(m.A)));
+end
+
+fprintf('\nSaved mat_fixture.mat and mat_two_model.mat. Next: gate_check.py.\n');
--- a/.context/issue-159/gate_prep.py
+++ b/.context/issue-159/gate_prep.py
@@ -1,0 +1,48 @@
+"""MATLAB gate prep for issue #159 (run first).
+
+Writes a real 2-model pyAMICA output, then dumps what Python ``loadmodout``
+reads for W/A/sbeta/rho on (a) the genuine single-model Fortran fixture and
+(b) that 2-model output. ``gate_compare.m`` then loads the same two directories
+with EEGLAB's ``loadmodout15.m`` and this script's companion assertion (run last)
+checks they agree element-wise. Python ``loadmodout`` is a port of
+``loadmodout15.m``, so on correct byte order the two must match to ~1e-12; the
+pre-#159 C-order read made W (and A/svar) disagree by the internal transpose.
+"""
+
+import os
+import numpy as np
+
+from pyAMICA import AMICA
+from pyAMICA.numpy_impl.load import loadmodout
+from pyAMICA.torch_impl.utils import load_eeglab_data
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+ROOT = os.path.abspath(os.path.join(HERE, "..", ".."))
+SAMPLE = os.path.join(ROOT, "pyAMICA", "sample_data")
+FIXTURE = os.path.join(SAMPLE, "amicaout")
+TWO_MODEL = os.path.join(HERE, "gate_2model")
+
+X = load_eeglab_data(os.path.join(SAMPLE, "eeglab_data.fdt"), 32, 30504).astype(
+    np.float64
+)
+
+# Deterministic real 2-model fit -> genuine-Fortran multi-model output.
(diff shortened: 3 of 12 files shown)


neuromechanist:
## PR review (4 Sonnet reviewers: code, tests, comments, silent-failure)

All findings addressed. Highlights:

**Critical (found independently by the comment and code reviewers, empirically confirmed): `compare_with_fortran`'s added `.T` was backwards.** The real caller (`examples/torch_example.py:79`) passes `get_unmixing_matrix(0)`, which is already the true/Fortran convention; since the #159 loader fix makes `out.W` the same convention, the two compare directly. My `.T` re-broke it (direct match |corr|=1.000 vs 0.985 with `.T`). Reverted; comment corrected. Note this function is uncalled in tests, so CI did not catch it.

**Test quality (test-analyzer, empirically verified against reconstructed `main`):**
- `test_loadmodout_reproduces_fortran_reported_ll` and `test_loadmodout_sources_reproduce_live_transform_multimodel` confirmed genuine (fail on pre-fix code).
- `test_load_results_returns_internal_w_multimodel` was shown to pass identically on `main` (it guards writer↔reader lockstep, not the original bug). Docstring corrected, and a **new direct byte-layout test** (`test_written_w_bytes_are_genuine_fortran_layout`) now pins the writer's genuine-Fortran multi-model layout in CI, independent of any reader.
- Overstated "0/32" figure in a docstring softened to what's reproducible.
- Added `share_comps` multi-model round-trip test and the missing `sources()` validation branches (negative model, 1-D X).

**Silent-failure:**
- Legacy multi-model directories: a format-version stamp is infeasible (genuine Fortran output has no marker and must be read with the new convention, so unmarked can't be refused). Added a **changelog migration note** instead; pre-fix multi-model `W` was never in the correct convention regardless.
- Added writer shape validation (`W` must be 3-D; model axis must match `gm`).
- Stale `_write_results` docstring (`core.py`) and the "each array" overreach in `write_amicaout` corrected.

**Out of scope, filed separately:** #164 (pre-existing `fit()` crash on `do_approx_sphere=False + pcakeep`, the only genuine `num_pcs < data_dim` path).

MATLAB gate unchanged (writer bytes unchanged): W/A/sbeta/rho agree with `loadmodout15.m` to ≤1.3e-13 single- and multi-model.

Issue #173, opened 2026-07-17
Native windows-arm64 AMICA binary (no MSYS2 gfortran yet)


neuromechanist:
Part of epic #165. The release matrix builds native binaries for macOS arm64, Linux x64/arm64 and Windows x64, but **windows-arm64 has no native Fortran toolchain**: MSYS2 CLANGARM64 ships clang (no `gcc-fortran`), and LLVM flang does not compile amica15's legacy F90 (needs `-std=legacy`/`-fallow-argument-mismatch`, which flang lacks). The workflow keeps windows-arm64 as an experimental (continue-on-error) job so it lights up automatically if a toolchain lands.

Interim: Windows 11 ARM runs x64 binaries via emulation, so the Phase 3 (#168) binary resolver maps windows-arm64 -> the windows-x64 asset. That covers Windows-ARM users functionally at an emulation performance cost.

## Acceptance
- A native windows-arm64 binary built and smoke-tested in CI (via a working gfortran-for-aarch64 MSYS2 package if/when available, or a flang build path with the source adjustments flang needs), attached to the release.
- Resolver prefers the native arm64 asset over x64-emulation once it exists.

Issue #165, opened 2026-07-17
Epic: native Fortran run engine + cross-platform binary distribution


neuromechanist:
Ship the native AMICA Fortran reference as a first-class run engine (the 4th, alongside NumPy, PyTorch and MLX), with prebuilt binaries for macOS, Linux x64, Linux ARM64, Windows x64 and Windows ARM64 attached to each GitHub release, and a package installer that resolves and copies the right binary for the host system.

## Context

The AMICA Fortran reference (`amica15.f90` + `funmod2.f90`) is currently only used in `validate_implementations.py` and bundled as a single x86 macOS binary (`sample_data/amica15mac`), which is not shipped in the wheel. sccn/amica PR #53 established the vendor-neutral build recipe we abide by: gfortran + Open MPI + LAPACK, no MKL, with a portable `random_seed` patch and `vmath_shim.c` (vrda_exp/vrda_log as libm loops). That recipe still dynamically links Open MPI + LAPACK at runtime and does not cover Windows.

## Key technical decision: single-rank MPI shim

`amica15.f90` uses ~15 MPI calls (BCAST, REDUCE, ALLREDUCE, BARRIER, COMM_SIZE/RANK, GATHER, COMM_SPLIT, INIT/FINALIZE), all trivial for a single rank (pyAMICA runs it as one process). Providing a tiny single-rank MPI shim (the same philosophy as `vmath_shim.c`, extended from no-MKL/no-AMD-LibM to no-MPI-runtime), plus static LAPACK/BLAS, yields dependency-free static binaries that run with no `brew`/`apt` install, on every target including Windows.

## Phases

1. Dependency-free single-rank build: MPI shim + static LAPACK; a portable build script producing a binary that runs with zero runtime deps, validated against the Fortran reference LL on real sample EEG.
2. Cross-platform release CI: GitHub Actions matrix building the five targets and attaching them as release assets (checksummed).
3. Native run engine + binary resolver/installer: a Python `AMICANative` backend that resolves the host binary (bundled or downloaded from the release), runs it, and parses output via `loadmodout`.
4. Package rename to `pAMICA`/`pamica` (PyPI `pyamica` is taken); consistent distribution + import naming.

## Acceptance

- `pip install`-ing the package on each of the five platforms yields a working native engine with no manual toolchain setup.
- The native engine reproduces the Fortran reference decomposition (parity via `loadmodout` + LL).
- Binaries are attached to the release, checksummed, and resolved automatically per host.

[...]

PR #175, opened 2026-07-17
Phase 3: native run engine + binary resolver/installer

### LINKED ISSUE #168: Phase 3: native run engine + per-system binary resolver/installer
Part of epic #165. A Python AMICANative backend (the 4th run engine) that resolves the host binary (bundled or downloaded from the release), runs it, and parses output via loadmodout; installer copies the right binary per system.

neuromechanist:
## Summary

Adds `AMICANative`, the fourth run engine (alongside NumPy, PyTorch, MLX). It runs the dependency-free native binary and returns an `AmicaOutput` with the usual accessors, so it is the parity oracle the Python backends are checked against.

Closes #168. Part of epic #165.

## What's included

- `pyAMICA.native.resolver`: maps the host to its release asset, downloads on first use with SHA-256 verification, caches it, and honors `PAMICA_NATIVE_BINARY`. windows-arm64 -> windows-x64 (emulation, #173). `python -m pyAMICA.native` installs the binary explicitly.
- `pyAMICA.native.AMICANative`: writes the data (`.fdt`, column-major) + a full `input.param` (embedded template, so no dependency on installed sample data), runs the binary, and loads the result via `loadmodout`. Friendly `n_models`/`n_mix` aliases; a collapsed fit is reported as a clear degenerate-fit error instead of an opaque SVD failure (cf. #50).
- Exported as `from pyAMICA import AMICANative`; `pyAMICA.native` added to the wheel packages.
- CI builds the binary on Linux and sets `PAMICA_NATIVE_BINARY`, so the engine's end-to-end tests run for real (not skipped).

## Test plan

- 15 tests: resolver logic (platform tag, asset alias, SHA-256 verify, override, no-download) + real end-to-end engine runs on sample EEG (single- and multi-model, transform, degenerate-fit detection, input validation). No mocks; the E2E tests use a real locally/CI-built binary.
- Full non-slow suite: 292 passed, 1 xfailed, no regressions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,16 @@ jobs:
           python-version: "3.12"
           enable-cache: true
       - run: uv sync
+      # Build the dependency-free native binary (epic #165) so the AMICANative
+      # engine's end-to-end tests run for real on Linux (they skip without
+      # PAMICA_NATIVE_BINARY). Static reference LAPACK/BLAS, same as the release.
+      - name: Build native AMICA binary (for engine E2E tests)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gfortran liblapack-dev libblas-dev
+          LAPACK_LIBS="-l:liblapack.a -l:libblas.a" EXTRA_LDFLAGS="-static-libquadmath" \
+            bash native/build.sh
+          echo "PAMICA_NATIVE_BINARY=$PWD/native/amica15_shim" >> "$GITHUB_ENV"
       # Exclude the "slow" parity tests: they invoke the macOS-only Fortran
       # reference binary (pyAMICA/sample_data/amica15mac), which cannot run on
       # the Linux runner. MPS-specific tests self-skip when MPS is absent.
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -5,6 +5,17 @@ Release notes are also published on the
 
 ## Unreleased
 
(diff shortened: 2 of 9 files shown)


neuromechanist:
## Review (Sonnet) addressed

**Critical (security): fixed.** The resolver downloaded straight to the final cache path and verified *after*, so a corrupt/tampered binary poisoned the cache and (on Windows, where `.exe` needs no exec bit) could execute unverified on later calls, plus a TOCTOU race. Now the download stages in a private temp dir on the same filesystem, is SHA-256-verified and chmod'd there, and only then atomically renamed into the cache path -- so a file exists at the cache path only once verified. Regression tests added: a bad checksum raises and leaves the cache path absent (no poisoning), and a subsequent good download still succeeds.

**Important: fixed.** A relative `binary`/`PAMICA_NATIVE_BINARY` path passed the existence check but failed at exec time (subprocess runs with `cwd=tempdir`); both now `Path(...).resolve()`.

**Minor: addressed.** `version` cache segment guarded against `../` traversal (`Path(version).name`); empty/truncated `W` now caught alongside the NaN degenerate-fit check; added a comment on the `files`-first param ordering requirement.

**Confirmed clean by the reviewer:** `_fmt` scientific-notation fidelity (byte-identical to `input.param`), URL host-injection safety, lazy top-level import (no network at import), and tempdir/output lifetime (all arrays materialized before cleanup).

17 native tests pass (incl. the 2 new security regression tests); lint/ty clean.

Issue #143, opened 2026-07-14
Phase 4: expose PMI/MIR through the MNE wrapper


neuromechanist:
Part of #139. Depends on Phase 1 and on the MIR/PMI epic (#133).

Once #133's MIR/PMI port lands, expose it through the MNE-facing wrapper too
(e.g. `ica.mir(raw)` / `ica.pmi()`), so MNE-side users get the same
separation-quality metrics EEGLAB-side users get via #133's phases.

neuromechanist:
Delivered in PR #190 (squash-merged into the epic branch, commit 8978514).

PR #190, opened 2026-07-18
Phase 4: expose PMI/MIR through the MNE wrapper

neuromechanist:
## Summary

Wires the #133 separation-quality metrics onto the MNE-facing wrapper: `AMICAICA.mir(inst, model_idx=...)` (Mutual Information Reduction, in nats) and `AMICAICA.pmi(inst, model_idx=...)` (pairwise mutual information between the fitted sources). Both extract the fitted channels from a `Raw`/`Epochs` and delegate to `AMICA.mir`/`pmi` (which already exist from #133), so MNE-side users get the same metrics as EEGLAB-side users.

Closes #143. Part of epic #139. This is the final phase.

## What's new

- `AMICAICA.mir(inst, *, model_idx=0, nbins=None)` and `pmi(inst, *, model_idx=0, nbins=None)`. `model_idx` is keyword-only (consistent with the other consumer methods) and validated against the fitted model count; both guard unfitted/degenerate via `_check_fitted`. `Epochs` are concatenated along time.

## Test plan

- MNE wrapper (`test-mne` job, real EEG): `mir`/`pmi` reproduce `AMICA.mir`/`pmi` on the picked-channel array exactly, per model; `pmi` shape is `(n_components, n_components)`; both work on `Epochs`; `model_idx` bounds and unfitted guards. Round-trip verified: wrapper MIR == array-API MIR (33.54 / 33.58 for the two models).
- Local gates green: ruff, `ty check .`, `mkdocs build --strict`, typos, and the mne suite (44) on real sample EEG.

### DIFF
--- a/docs/api/mne-compat.md
+++ b/docs/api/mne-compat.md
@@ -109,4 +109,17 @@ generalized-Gaussian shape (meaningful for `pdftype=0`). `shared_components`
 lists components merged across models by `share_comps` (empty otherwise). The
 same three accessors exist on the scikit-learn-style [`AMICA`](amica.md).
 
+## Separation-quality metrics
+
+The [MIR/PMI metrics](metrics.md) are available directly on an MNE object, so
+MNE-side users get the same separation-quality numbers as EEGLAB-side users:
+
+```python
+mir_nats, variance = ica.mir(raw, model_idx=0)   # mutual information reduced
+mi_matrix = ica.pmi(raw, model_idx=0)            # pairwise MI between sources
+```
+
+Both extract the fitted channels from the `Raw`/`Epochs` and delegate to
+`AMICA.mir`/`pmi`, reproducing the array API exactly.
+
 ::: pamica.mne_compat.AMICAICA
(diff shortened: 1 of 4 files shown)


neuromechanist:
## Review (2 Sonnet reviewers: code, tests)

Both confirmed the phase is thin, correct wiring; the code reviewer found nothing above its confidence bar (traced the full `AMICAICA` -> `AMICA` -> `AMICATorchNG` delegation chain, ran all 44 tests, verified the round trip). Test reviewer confirmed the per-model comparison genuinely discriminates (model 0 MIR 33.46 vs model 1 33.55). Actionable findings addressed:

- **`pmi` on Epochs was only shape-checked**, not value-pinned, so a bug in `_data_for`'s Epochs branch would slip through for `pmi` (caught for `mir`). Now value-pinned against `AMICA.pmi`.
- Added an **`nbins` passthrough test** (matches `AMICA` with a non-default `nbins`, so a dropped forward is caught) and a **degenerate-fit refusal test** for `mir`/`pmi` (matching the convention used for the other consumer groups).
- Made **per-model discrimination explicit** (`assert not allclose(model0, model1)`), self-documenting the model_idx-honored invariant.
- Added `mir`'s missing return annotation.

Local gates green: ruff, `ty check .`, `mkdocs build --strict`, typos, and the mne suite (46) on real sample EEG.

Issue #142, opened 2026-07-14
Phase 3: preserve pyAMICA-specific metadata through the MNE wrapper


neuromechanist:
Part of #139. Depends on Phase 1.

Adaptive source-density family (`pdftype` 0-4) per component, the GG shape
parameter `rho`, and component-sharing state should remain inspectable through
the MNE-facing objects, not silently dropped in the name of MNE compatibility.

neuromechanist:
Delivered in PR #189 (squash-merged into the epic branch, commit 81e1f45).

PR #191, opened 2026-07-18
MNE-Python compatibility layer (epic #139)

### LINKED ISSUE #139: Epic: MNE-Python compatibility layer (additive, EEGLAB I/O unchanged)

neuromechanist:
## Summary

Adds an MNE-Python compatibility layer for pamica, additive throughout: the scikit-learn-style `AMICA` API and the byte-identical EEGLAB I/O are unchanged. MNE-Python users get a native entry point via `from pamica.mne_compat import AMICAICA`, shipped as the optional `pamica[mne]` extra so `import pamica` never requires MNE. Works with current MNE, no fork or extension.

Closes #139.

## Phases

- **Phase 1 (#140):** core single-model wrapper. `AMICAICA.fit(raw/epochs, picks=...)` and `to_mne_ica()` mapping pamica's mean/symmetric-ZCA sphere/unmixing into a real `mne.preprocessing.ICA` (round-trips `AMICA.transform` to float64 precision); `get_sources`/`apply`/`get_components`/`plot_components`/`plot_sources` delegate to it.
- **Phase 2 (#141):** multi-model exposure. Each model exports as its own `mne.preprocessing.ICA` (`to_mne_ica(model_idx=...)`); the per-sample model dominance MNE cannot represent is exposed via `get_model_probability`/`plot_model_probability`, built on a new public `AMICA.model_loglik`/`model_probability` accessor (validated bit-for-bit against the E-step's `Lht`).
- **Phase 3 (#142):** pamica-specific metadata (`get_pdftype`/`get_rho`/`shared_components` + `PDFTYPE_NAMES`) surfaced rather than dropped by the ICA export.
- **Phase 4 (#143):** the #133 separation-quality metrics (`mir`/`pmi`) callable directly on an MNE object.

## Design notes

- MNE has no multi-model concept, so multi-model AMICA is exposed per-model plus a dominance accessor, not forced into MNE's single-unmixing `ICA`.
- The per-model export folds each model's data-space center `c` into `pca_mean_`; single-model (`c=0`) stays byte-identical.
- `pca_components_` is kept orthonormal (sphere written as `V diag(1/sqrt(e)) V^T`) so MNE's scalp maps come out in channel space.
- PCA reduction (`pcakeep`/`pcadb`) is rejected (the export assumes full-rank whitening); non-finite input and degenerate fits are refused.

## Test plan

- New `pamica/tests/mne_tests/` (46 tests) run on the real bundled `eeglab_data.set` via a dedicated `test-mne` CI job (`uv sync --extra mne`); the base job skips them via `importorskip`. New backend suites (`test_ng_model_posterior.py`, `test_ng_metadata.py`) run in base CI. No mocks.
- Every phase went through a 4-5 reviewer Sonnet pass; all findings addressed (see the per-phase PRs #187-#190). CI green on all four.

## Merge

Per the repo policy, an epic PR uses a regular merge (not squash) to preserve the per-phase history.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,6 +113,28 @@ jobs:
           files: coverage.xml
           fail_ci_if_error: false
 
+  test-mne:
+    name: Test MNE wrapper (optional [mne] extra)
+    needs: [lint, typecheck]
+    runs-on: ubuntu-latest
+    env:
+      PYTORCH_ENABLE_MPS_FALLBACK: "1"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.12"
+          enable-cache: true
+      # mne IS pip-installable on Linux (unlike the Apple-only mlx extra), so the
+      # AMICAICA wrapper (issue #139) is exercised for real in CI here. Its tests
+      # live in pamica/tests/mne_tests and self-skip in the base `test` job
+      # (importorskip on mne); this job installs the [mne] extra so they run on
+      # the real sample EEG. pamica/mne_compat is omitted from coverage in
+      # pyproject (the base env has no mne), so no fail-under gate applies here.
+      - run: uv sync --extra mne
+      - name: pytest (mne wrapper, real sample EEG)
+        run: uv run pytest pamica/tests/mne_tests
+
   build:
     name: Build + import (Python ${{ matrix.python-version }})
     needs: [lint, typecheck]
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -27,3 +27,13 @@ from pamica.mlx_impl import AMICAMLXNG  # requires the `mlx` extra
 
 - **[`AMICAMLXNG`](mlx-backend.md)** — the optional Apple-GPU (MLX) backend; the
   fastest option on Apple Silicon (float32).
+
+The optional MNE-Python wrapper is likewise imported explicitly:
+
+```python
+from pamica.mne_compat import AMICAICA  # requires the `mne` extra
+```
+
+- **[`AMICAICA`](mne-compat.md)** — fit AMICA from an MNE `Raw`/`Epochs` and
+  interoperate with `mne.preprocessing.ICA` (`get_sources`, `apply`,
+  `plot_components`, `to_mne_ica`).
--- a/docs/api/mne-compat.md
+++ b/docs/api/mne-compat.md
@@ -1,0 +1,67 @@
+# MNE-Python compatibility (AMICAICA)
+
+`AMICAICA` fits AMICA directly from an [MNE-Python](https://mne.tools)
+`Raw`/`Epochs` and hands the result back through the standard MNE ICA surface.
+It is **additive**: the scikit-learn-style [`AMICA`](amica.md) interface and the
+byte-identical [EEGLAB output](../guides/eeglab.md) are unchanged; this is a
+second entry point for MNE users, not a replacement.
+
+MNE is an optional dependency, so `import pamica` never requires it. Install the
+extra and import the wrapper explicitly:
+
+```bash
+pip install pamica[mne]
+```
+
+```python
+import mne
+from pamica.mne_compat import AMICAICA
+
+raw = mne.io.read_raw_eeglab("subject.set", preload=True)
+
+ica = AMICAICA(n_mix=3, random_state=42).fit(raw, picks="eeg", max_iter=100)
+
+sources = ica.get_sources(raw)     # an mne.io.RawArray of component activations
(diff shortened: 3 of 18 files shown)


neuromechanist:
## Integrated review (Sonnet, cross-phase)

Ran a focused review over the combined 4-phase diff for integration/coherence issues (beyond the per-phase reviews on #187-#190). The layering checked out: guard ordering (`_check_fitted` before `_check_model_idx`) is uniform across all 13 call sites; the two same-named `_check_model_idx` validators (wrapper + backend, added in different phases) produce identical errors and bound against the same value (redundant defense-in-depth, not a double-validation bug); the phase-3 backend `_check_model_idx` retrofit breaks no existing caller; the `viz.plot_model_probability` dual `out`/`lht` API is only ever called with `lht=` from the wrapper; packaging (`tool.setuptools.packages`), the `test-mne` CI job + coverage omit, and the docs/changelog are all complete across the four phases.

Two findings:

- **Module docstring was stale** (labeled "phase 1", omitted phases 3-4). Fixed in the follow-up commit — it now covers the metadata accessors and mir/pmi.
- **`to_mne_ica`'s `model_idx` is not keyword-only** unlike the other model-scoped methods. Kept as positional-or-keyword by design: the keyword-only convention on `get_sources`/`apply`/`plot_components`/etc. exists specifically to protect their `*args`/`**kwargs` passthrough to MNE from a positional collision. `to_mne_ica` takes only `model_idx` (no passthrough, no collision), so `to_mne_ica(h)` is the natural call and is what the internal delegators and tests use; forcing keyword-only would break those for no safety benefit.

CI re-running after the docstring commit.

Issue #193, opened 2026-07-18
rholrate decays permanently instead of resetting per iteration (Fortran parity)


neuromechanist:
## Summary

Found while investigating #145. `AMICATorchNG`'s rho learning-rate schedule diverges from the Fortran reference: **torch permanently decays `rholrate` on every log-likelihood decrease**, whereas Fortran **resets `rholrate = rholrate0` each iteration** and only ratchets the ceiling at `maxdecs` (amica15.f90:1788). This is a genuine Fortran-parity discrepancy in the rho (generalized-Gaussian shape) update, independent of #145.

## Detail

- Fortran: each iteration starts from `rholrate0`; the anneal only tightens a ceiling after `maxdecs` consecutive LL decreases.
- torch (`pamica/torch_impl/core.py`): once an LL decrease shrinks `rholrate`, it stays shrunk for the rest of the fit (monotone decay), so the rho update is progressively under-relaxed relative to Fortran.

## Caveat (why it was not fixed under #145)

Fixing this to match Fortran actually makes the #145 long-budget Newton parity metric *worse* on the (data-inadequate) bundled recording — the accidental decay freezes rho and removes a Newton noise source. That is a red herring for #145; this issue is about **correctness/parity of the rho schedule on its own merits**, and should be evaluated against the Fortran binary as the spec (not against the #145 metric).

## Acceptance

Match Fortran's per-iteration `rholrate` reset + ceiling-ratchet behavior; validate the rho trajectory / converged parity against the `amica15mac` binary on the bundled sample data. Mirror into `numpy_impl/core.py` if the same defect is present there.

Refs #145.

neuromechanist:
Fixed by #194 (a6b967f): rholrate now resets to rholrate0 each iteration with the maxdecs ceiling ratchet, matching amica15.f90. Same bug in MLX tracked in #195.

Issue #145, opened 2026-07-15
AMICATorchNG diverges from Fortran on weak components at 2000 iters, despite Newton making the optimum sharply unique


neuromechanist:
## Background

Follow-up to #144. While re-verifying Table 1's single-model correlation
claim on a properly data-adequate real recording (OpenNeuro ds002718
sub-002, 70 channels, full 747,750 frames, k=152.6 -- well above the
project's own k>=60 threshold, #90), a consistent pattern emerged across
every dataset tested (the original bundled 32ch/30504-frame data at k=29.8,
a 32ch/147k-frame ds002718 subset at k=144, and the full 70ch recording at
k=152.6): **short iteration budgets (200-300) reliably give 0.98-0.999
Fortran-vs-AMICATorchNG correlation; the full 2000-iteration budget (`do_newton=True`
in both) consistently gives lower, more variable correlation (0.87-0.97
mean), typically driven by one or two components dropping to 0.5-0.6 while
the rest stay well-matched.**

## This is not inherent data ambiguity -- it's an NG-specific divergence

Two independent Fortran fits of the *same* data (different random inits --
confirmed the randomness is real: `amica15_header.f90:97` hardcodes
`seed(2) = (123456, 654321)`, never read from `input.param`, but
`amica15.f90:219-220` mixes in `system_clock(c1)`, so the effective RNG
state differs on every invocation) agree with **each other** at
mean=0.9997, min=0.9983 (all 70/70 components >0.998) at 2000 iterations,
`do_newton=1`. If the low components were inherent non-identifiability
(as in the genuinely non-identifiable multi-model case), independent
Fortran runs should show the same spread on their weak components. They
don't -- Fortran is extremely self-consistent.

This within-Fortran number (0.9997) is *higher* than the project's own
prior k=152 cross-backend sweep (`.context/issue-90/ksweep_findings.md`:
native-fortran-f64 vs torch-cuda-f64 at 0.995, mean cross-backend 0.9817)
-- but that sweep explicitly ran with **`do_newton=0`** on every backend
(`benchmark_decompose.py`'s `_write_fortran_param`/`_fit_torch`, forced off
for benchmark fairness). Newton's quadratic convergence near an optimum
likely sharpens/resolves directions that plain natural gradient leaves
loose, which is consistent with both observations: without Newton, even
same-algorithm cross-device agreement plateaus at ~0.98; with Newton,
independent Fortran random inits converge to essentially the same sharp
optimum (0.9997).

[...]

PR #194, opened 2026-07-18
Reset rholrate per iteration to match Fortran

neuromechanist:
## Summary

`AMICATorchNG` (and the legacy `numpy_impl` backend) diverged from the Fortran reference in the rho (generalized-Gaussian shape) learning-rate schedule: **torch permanently decayed `rholrate` on every log-likelihood decrease**, whereas Fortran **resets `rholrate = rholrate0` each iteration** and only ratchets the ceiling at `maxdecs`. Once an LL decrease shrank `rholrate` it stayed shrunk (monotone decay), so on a long run the rho update was progressively under-relaxed and eventually frozen (rate collapsed to ~1e-5 within a few hundred iterations). Refs #193.

## Fortran expressions

- `amica15.f90:1788`/`:1795` — `update_params` sets `rholrate = rholrate0` every iteration, before the rho update at `:1814`. So Fortran's per-decrease `rholrate = rholrate*rholratefact` (`:1045`) is overwritten and never reaches the rho update; the effective rate is always the `rholrate0` ceiling.
- `amica15.f90:1050` — the ceiling `rholrate0` ratchets `*= rholratefact` only at `maxdecs`, gated on `iter > newt_start` (unlike `newtrate`, which is additionally gated on `do_newton`).

## Fix

rho has no per-iteration ramp, so the effective rho rate is exactly its ceiling. Both backends now carry that ceiling directly:
- **torch** (`pamica/torch_impl/core.py`): `self.rholrate` is the ceiling — reset to `rholrate0` at fit start (existing), ratcheted only in the `maxdecs` block (gated `it > newt_start`), never per-decrease. The buggy per-decrease `self.rholrate *= self.rholratefact` is removed.
- **numpy** (`pamica/numpy_impl/core.py`): same — `self.rholrate` is the ceiling, reset to `rholrate0` at fit start, ratcheted at `maxdecs` (the previously-dead `rholrate0` ratchet is retargeted to `rholrate`); the per-decrease decay is removed.

`rholrate0` stays the pristine constructor value in both. No state-dict format change (the persisted `rholrate` field now carries the ceiling; round-trip is unchanged).

## Parity validation (vs `amica15mac`, bundled 32ch/30504 data, 2000 iters, do_newton)

Converged rho distribution (shape param, bounded [1,2]) — sorted, permutation-free:

| | min | q25 | med | q75 | max |
|---|---|---|---|---|---|
| torch (fixed) | 1.000 | 1.606 | 1.822 | 1.994 | 2.000 |
| Fortran run #1 | 1.000 | 1.614 | 1.829 | 2.000 | 2.000 |
| Fortran run #2 | 1.000 | 1.636 | 1.845 | 2.000 | 2.000 |

Sorted-rho L1 distance: torch-vs-Fortran = 0.032, Fortran-vs-Fortran = 0.018 — torch's rho distribution sits within Fortran's own run-to-run spread on this data. Post-fit `rholrate` ends at 0.00039 (= `0.05 * 0.5**7`, seven `maxdecs` ratchets, matching `newtrate`), not the buggy per-decrease collapse (`0.05 * 0.5**19 ~ 1e-7`).

## Note on #145

This is a correctness/parity fix on its own merits; it is evaluated against the Fortran binary, NOT the #145 long-budget Newton metric. As documented in #193, matching Fortran here **worsens** the #145 min-correlation on the (data-inadequate, k=29.8) bundled recording, because the accidental decay was freezing rho and removing a Newton noise source. That is expected and is a red herring for #145 (see the #145 investigation notes).

## Test plan

- New real-data regression test `test_rholrate_ratchets_at_maxdecs_not_per_decrease` (marked slow): on the sample EEG a 300-iter Newton run overshoots and triggers several LL decreases; asserts the surviving `rholrate` ratcheted exactly as often as `newtrate` (both gated on `iter > newt_start` at `maxdecs`), and is orders of magnitude above the old per-decrease-decay value.
- Existing gates green: `test_full_fit_parity_numpy_vs_ng`, `test_end_to_end_correlation_vs_fortran`, `test_sample_data.py` (numpy-vs-Fortran), torch/numpy non-slow suites, `ruff`, `ty`.

Refs #193

### DIFF
--- a/pamica/numpy_impl/core.py
+++ b/pamica/numpy_impl/core.py
@@ -628,6 +628,9 @@ class AMICA:
             self.good_idx = saved_good_idx
             self.num_good_samples = saved_num_good
         self.lrate = self.lrate0
+        # rholrate is the (maxdecs-ratcheted) rho-rate ceiling; restore it to the
+        # pristine rholrate0 so a re-fit starts fresh (issue #193).
+        self.rholrate = self.rholrate0
         self.ll = []
         self.nd = []
 
@@ -1468,10 +1471,13 @@ class AMICA:
 
         Mirrors Fortran's per-iteration convergence handling
         (amica17.f90:1062-1103). On a likelihood decrease Fortran does NOT stop
-        at ``maxdecs``; it lowers the learning-rate *ceiling* (``lrate0``, and
-        ``newtrate``/``rholrate0`` under Newton) and continues, which is what
-        keeps a long run from oscillating and drifting past its converged
-        solution (issue #41). The updated ``numdecs``/``numincs`` counters are
+        at ``maxdecs``; it lowers the learning-rate *ceilings* (``lrate0``; the
+        rho rate once ``iter > newt_start``; ``newtrate`` under Newton) and
+        continues, which is what keeps a long run from oscillating and drifting
+        past its converged solution (issue #41). The rho ceiling is
+        ``self.rholrate`` here (reset to ``rholrate0`` each fit), ratcheted only
+        at ``maxdecs`` -- never per-decrease (issue #193). The updated
+        ``numdecs``/``numincs`` counters are
         returned so they accumulate across iterations (they previously did not).
 
         Parameters
@@ -1517,8 +1523,8 @@ class AMICA:
         grad_norm = self.nd[-1] if len(self.nd) > 0 else None
 
         # Likelihood decrease (Fortran amica17.f90:1062-1083): reduce the current
-        # lrate, and once maxdecs decreases have accrued, ratchet the ceiling
-        # (lrate0, plus newtrate/rholrate0 under Newton) down and continue --
+        # lrate, and once maxdecs decreases have accrued, ratchet the ceilings
+        # (lrate0; the rho rate; newtrate under Newton) down and continue --
         # NOT stop. Only a lrate/gradient floor terminates on a decrease.
         if self.ll[-1] < self.ll[-2]:
             if self.lrate <= self.minlrate or (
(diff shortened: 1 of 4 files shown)


neuromechanist:
## Review (/review-pr, reviewers = Sonnet)

Ran the pr-review-toolkit code-reviewer and pr-test-analyzer on the diff. Correctness checks were clean (ceiling-ratchet gating `iter > newt_start` correct and distinct from the do_newton-gated newtrate ratchet; `rholrate0` stays pristine; torch state_dict round-trip unaffected; cross-backend consistency confirmed). Findings and resolutions:

**1. numpy fix had no test that would fail without it (test-analyzer, sev 8) — ADDRESSED.** Extended the existing fast, deterministic unit test `test_check_convergence_ratchets_lrate_on_decrease` to assert the rho ceiling: it stays untouched on a decrease below `max_decs` (pre-#193 it decayed on the first decrease) and ratchets by `rholratefact` when `numdecs` hits `max_decs`. Fails on pre-fix numpy, no real data needed. (commit 7bae7f0)

**2. numpy `_reinitialize_for_restart` rho reset was new + untested (test-analyzer, sev 7) — ADDRESSED.** Added `test_reinitialize_for_restart_resets_rho_ceiling`: a direct unit test that a ratcheted `rholrate` is restored to `rholrate0` across a restart, mirroring the existing lrate reset. (commit 9dd3141)

**3. MLX backend has the same bug (code-reviewer) — TRACKED as #195, not folded in.** `mlx_impl/core.py:616` still decays `rholrate` per-decrease with no reset/ceiling. Out of #193's stated torch+numpy scope, and MLX is natural-gradient-only with no `newt_start`/Newton concept, so the Fortran ceiling-ratchet gate (`iter > newt_start`) needs a design decision. Filed #195 with the fix sketch; MLX is explicitly not a float64-parity backend, so it does not gate this PR.

**4. torch fit-start refit reset has no direct test (test-analyzer, sev 4) — pre-existing, out of scope.** `torch_impl/core.py:782` (`self.rholrate = self.rholrate0`) is unchanged by this PR and mirrors the existing lrate/newtrate/lrate_cap fit-start resets; noting as a possible follow-up rather than expanding this diff.

All added tests + existing parity gates (`test_full_fit_parity_numpy_vs_ng`, `test_end_to_end_correlation_vs_fortran`, `test_sample_data` numpy-vs-Fortran, torch/numpy suites, ruff, ty) are green.

Issue #221, opened 2026-08-15
Support for Maxwell-filtered Elekta MEG data


sunyuhongwr:
## Background
First of all, I think pAMICA is an excellent Python package. It provides a very comprehensive implementation of AMICA, and I really appreciate the amount of work and thought that has gone into developing it.
However, as far as I understand, pAMICA currently does not explicitly support MEG data, particularly Maxwell-filtered Elekta/MEGIN MEG data. I am currently trying to apply AMICA to my MEG data, and I encountered two issues that seem particularly relevant for MEG. A common preprocessing step for Elekta/MEGIN MEG data is Maxwell filtering. After Maxwell filtering, the resulting MEG data are generally not full-rank. For example, my original Elekta MEG data contain 306 MEG channels (102 magnetometers + 204 gradiometers), but after Maxwell filtering the numerical rank of the MEG data is only around 70. Therefore, the rank deficiency is expected as a consequence of Maxwell filtering and does not necessarily indicate a problem with the data.

## 1. Support for non-full-rank Maxwell-filtered MEG data
When I try to use pAMICA directly on Maxwell-filtered MEG data, the fit can become degenerate or the resulting decomposition cannot be directly mapped back to the original sensor space. This makes it difficult to use pAMICA directly with Maxwell-filtered MEG.
Would it be possible to support non-full-rank MEG data explicitly in pAMICA?

For example, one possible approach could be:
1. Estimate the numerical rank of the input MEG data.
2. Perform SVD/PCA reduction to the numerical rank. 
3. Represent the data in this reduced-rank space.
4. Run AMICA in the reduced-rank space.
5. Transform the AMICA decomposition back to the original MEG sensor space.
Conceptually, if the original data are
    X ∈ R^(306 × T)
and the numerical rank is approximately 70, I would construct a rank-70 representation
    X ≈ U_70 Z
where
    U_70 ∈ R^(306 × 70)
    Z   ∈ R^(70 × T).
AMICA would then operate on Z rather than directly on the 306-channel rank-deficient data. Afterwards, the decomposition could potentially be mapped back to the 306-channel sensor space through U_70.
Would this be a mathematically valid way of applying AMICA to Maxwell-filtered MEG data? 

## 2. Different scales of Elekta MEG channel types
[...]

PR #224, opened 2026-08-15
Handle rank-deficient input across all backends

### LINKED ISSUE #223: Rank-deficient input: no mineig detection, model dimension not reduced to numeigs
## Summary

pamica does not implement the Fortran reference's rank-deficiency handling. `amica15.f90` detects the numerical rank of the data covariance, reduces the model to that rank, and keeps a pseudo-inverse for mapping components back to sensor space. pamica does none of the three, so rank-deficient input (Maxwell-filtered MEG, average-referenced or interpolated EEG) either dies with `nan_ll` or silently fits a full-rank model to rank-deficient data.

Reported by an external user applying pamica to Maxwell-filtered Elekta MEG (306 channels, numerical rank ~70) in #221.

## What the Fortran does

Three pieces, all in `pamica/amica15.f90`:

1. **Automatic rank detection** (line 395):
   ```fortran
   numeigs = min(pcakeep, count(eigs > mineig))
   ```
   `mineig = 1.0e-15` by default (`amica15_header.f90:66`). This fires regardless of `pcakeep`, so rank deficiency is caught even with default settings.

2. **Model dimension follows the kept rank** (line 545):
   ```fortran
   nw = numeigs
   ```
   AMICA then estimates `numeigs` sources, not `nx`.

3. **Pseudo-inverse of the sphere for back-projection** (lines 550-560): `Spinv` is `(nx, numeigs)`, computed by SVD, and maps the decomposition back to the original `nx` sensor space.

## What pamica does

- No `mineig` equivalent anywhere (`grep -rn mineig pamica/*.py` is empty).
- `n_channels` is taken from the input row count (`amica.py:207`) and passed to the backend (`amica.py:225`); `n_comps = n_channels * n_models` (`torch_impl/core.py:541`). Nothing ever resizes it to the kept rank.
- No `Spinv`. `get_mixing_matrix()` returns A in the sphered space, and callers must de-sphere by hand.

## Reproducer

Real sample EEG projected onto its top-20 subspace (what SSS does to MEG):

```python
X = load_eeglab_data("pamica/sample_data/eeglab_data.fdt", 32, 30504, dtype=np.float32).astype(np.float64)
X = X - X.mean(1, keepdims=True)
U = np.linalg.svd(X, full_matrices=False)[0]
X_rd = U[:, :20] @ (U[:, :20].T @ X)     # 32 channels, rank 20

AMICA(verbose=False).fit(X_rd, max_iter=30, seed=0)
# converged_=False  stop_reason_='nan_ll'  at iteration 0
```

The symmetric ZCA sphere `V diag(1/sqrt(lambda)) V^T` inverts `sqrt(lambda ~ 0)` in the null directions. The `nan_ll` stop (#50) correctly refuses the model, but the Fortran would never have got here: it would have set `numeigs = 20` and fitted a 20-source model.

Current `pcakeep`/`pcadb` do not cover this:

- `pcakeep=20` alone keeps the symmetric ZCA form, so the sphere stays `(32, 32)` at rank 20 and AMICA still estimates 32 components from rank-20 data. No NaN, but 12 components are spurious.
- `pcakeep=20, do_approx_sphere=False` is the genuinely dimension-reducing path and **crashes**:
  ```
  RuntimeError: The size of tensor a (20) must match the size of tensor b (32) at non-singleton dimension 0
  ```
  `_preprocess` returns `(20, T)` from `sphere = inv_sqrt @ V.T` while every parameter is sized 32.

## Proposed fix

1. Port `mineig` (default `1e-15`) and compute `numeigs = min(pcakeep, count(eigs > mineig))` in `_preprocess`.
2. Set the model dimension from `numeigs` rather than `X.shape[0]`, so `n_comps` follows the kept rank. This fixes the `do_approx_sphere=False` crash as a side effect.
3. Carry `Spinv` on the model and have `get_mixing_matrix()` return sensor-space maps when reduction is active, instead of requiring callers to de-sphere manually.

Note for step 1: `mineig` is an **absolute** threshold on covariance eigenvalues, so it is unit-dependent. EEG in microvolts gives eigenvalues of order 1-100 and the default works; MEG in Tesla gives ~1e-26 and every eigenvalue falls below `1e-15`, which would collapse `numeigs` to 0. Matching Fortran exactly means inheriting that. Worth considering a relative fallback (`eigs > mineig * eigs.max()`) alongside the parity-faithful absolute one, and it is a further argument for channel-type scaling before sphering (#221).

## Related

- #221 (the MEG report that surfaced this)
- #50 (the degenerate-fit contract that catches the NaN today)

neuromechanist:
Closes #223.

Ports the Fortran reference's numerical-rank handling, which pamica had dropped. Rank-deficient input (Maxwell-filtered MEG, average-referenced or interpolated EEG) previously died with `nan_ll` on iteration 0, or silently fitted a full-rank model to rank-deficient data. Reported in #221 by a user applying pamica to 306-channel Elekta MEG at numerical rank ~70.

## What Fortran does, and now we do

| Fortran | Reference | pamica before | pamica now |
|---|---|---|---|
| `numeigs = min(pcakeep, count(eigs > mineig))` | amica15.f90:395 | absent | `mineig`, default `1e-15` |
| `nw = numeigs` | amica15.f90:545 | `n_channels` fixed from `X.shape[0]` | model sized to kept rank |
| `Spinv` (nx, numeigs) | amica15.f90:550-560 | absent | `get_sensor_mixing_matrix()` |

The reduced sphere follows Fortran's construction exactly, including the `do_approx_sphere` branch that symmetrizes the reduced whitening by the orthogonal polar factor of the leading `numeigs` block of `V^T` (amica15.f90:483-490).

## Parity

Full-rank data keep every eigenvalue, so the reduction path never fires and the existing code paths are untouched. Verified **bit-identical** before/after on the bundled sample EEG across single-model, Newton-enabled, and two-model fits (log-likelihood, sphere sum, mixing-matrix sum, and the first five LL-history entries all compare equal by `repr`).

## Improving on the reference

`mineig` is an *absolute* eigenvalue floor, so it is unit-dependent. Two consequences, both handled:

1. **MEG in Tesla** has covariance eigenvalues ~1e-26, all below `1e-15`. Fortran would compute `numeigs = 0` and proceed; pamica raises a `ValueError` naming the cause and the fix rather than fitting an empty model.
2. **The absolute floor is a poor rank detector**, because numerical-zero eigenvalues straddle it. On real EEG projected to rank 20:

   | threshold | rank found | reconstruction error |
   |---|---|---|
   | `mineig=1e-15` (Fortran default) | 24 | 1.98e-09 |
   | `mineig_rel=1e-9` | 20 (exact) | 7.52e-15 |

`mineig_rel` is opt-in and replaces the absolute floor when set, so the default stays Fortran-exact while the better detector is one argument away.

## Also fixed

`pcakeep=N, do_approx_sphere=False` — the genuinely dimension-reducing path — raised `RuntimeError: The size of tensor a (N) must match the size of tensor b (n_channels)`. `_preprocess` returned `(N, T)` while every parameter was still sized to the input channel count. Covered by a regression test.

## Tests

`pamica/tests/torch_tests/test_ng_rank_deficient.py`, 7 tests, real sample EEG only. Rank deficiency is produced by projecting the real recording onto its own leading subspace — the same rank-reducing linear operator SSS applies to MEG — so no synthetic data is involved.

## Follow-ups (not in this PR)

- `numpy_impl` and `mlx_impl` still lack rank handling.
- `mne_compat` sets a unit pre-whitener (`ica.pre_whitener_ = np.ones((n_ch, 1))`), so mixed magnetometer/gradiometer data are not channel-type scaled. Since the scaling determines which directions survive rank truncation, this matters for MEG; tracked in #221.

### DIFF
--- a/.context/decisions/0004-rank-deficient-input-handling.md
+++ b/.context/decisions/0004-rank-deficient-input-handling.md
@@ -1,0 +1,96 @@
+# ADR 0004: Rank-deficient input handling and a relative eigenvalue floor
+
+**Status:** accepted
+**Date:** 2026-08-15
+**Owner:** Seyed Yahya Shirazi
+
+## Context
+
+Rank-deficient input is ordinary in EEG/MEG, not exotic: Maxwell filtering leaves
+306-channel Elekta data at numerical rank ~70, average referencing removes exactly one
+dimension, and channel interpolation removes one per interpolated channel. The Fortran
+reference handles this and pamica did not, which issue #221 surfaced from a real MEG
+analysis.
+
+Fortran detects the rank, sizes the model to it, and keeps a pseudo-inverse to map
+components back to sensors:
+
+```fortran
+numeigs = min(pcakeep, count(eigs > mineig))   ! amica15.f90:395, mineig = 1e-15
+nw = numeigs                                   ! amica15.f90:545
+allocate(Spinv(nx,numeigs))                    ! amica15.f90:550-560
+```
+
+pamica had none of the three. The symmetric ZCA sphere therefore inverted
+`sqrt(lambda ~ 0)` and the fit died at `nan_ll` on iteration 0 (issue #223).
+
+`mineig` is an **absolute** floor on covariance eigenvalues, which makes it
+unit-dependent. Two consequences measured on the bundled sample EEG:
+
+- MEG in Tesla has eigenvalues ~1e-26, entirely below 1e-15. Fortran computes
+  `numeigs = 0` and proceeds.
+- Average-referenced EEG lands at `lambda_min/lambda_max = 8.5e-17`, so whether the
+  zero dimension is detected depends on the recording's absolute scale. It is decided
+  by luck.
+
+## Decision
+
+Port all three pieces of the Fortran mechanism, and **default to a relative floor**:
+`mineig_rel = 1e-12`, applied as `mineig_rel * largest_eigenvalue`. Setting
+`mineig_rel=None` restores Fortran's absolute-only behavior exactly.
+
+The rank decision lives in `pamica/rank.py` and is called by all three array backends,
+so they cannot drift (`.rules/backend_parity.md`).
+
+`mineig_rel` **replaces** the absolute floor rather than combining with it. Taking the
+larger of the two was tried first and is wrong: a relative floor for Tesla-scale data is
+~1e-35 in absolute terms, so `max()` silently discards it and reproduces the bug.
+
+## Consequences
+
+**Parity is unaffected for well-conditioned data.** Real EEG has
+`lambda_min/lambda_max ~ 5e-4`, eight orders above the relative floor, so nothing is
+reduced and results are bit-identical (verified against the pre-change implementation
+across single-model, Newton, and two-model fits).
+
+**Results change for rank-deficient data, in the intended direction.** On real EEG
+projected to rank 20:
+
+| threshold | rank found | reconstruction error |
+|---|---|---|
+| `mineig=1e-15` (Fortran) | 24 | 1.98e-09 |
+| `mineig_rel=1e-12` (default) | 20 | 7.52e-15 |
+
+**This is a deliberate divergence from the reference**, the first where pamica's default
+is not what Fortran does. It is recorded in `docs/guides/amica-differences.md` so it
+cannot be mistaken for a parity defect.
+
+**Average-referenced EEG now yields n-1 components rather than n.** Correct, but a
+visible behavior change for existing users, since the n-th component was previously
+numerical noise from a singular sphere.
+
+**A new obligation:** `get_sensor_mixing_matrix()` is now the way to obtain scalp maps.
+`get_mixing_matrix()` still returns the sphered-space `A`, which is not projectable to
+sensors when the sphere is non-square.
+
+## Alternatives considered
+
+- **Absolute floor only, matching Fortran exactly.** Loses on MEG (rank 0) and makes
+  average-reference detection scale-dependent. Rejected: bug-compatibility with the
+  reference is not the goal, numerical agreement on valid data is.
+- **`max(mineig, mineig_rel * lambda_max)`.** Implemented first, then measured: the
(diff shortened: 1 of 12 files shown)


neuromechanist:
## Update: folded backend parity and documentation in

Per review direction, this is no longer a PyTorch-only change.

### All backends, one shared policy
Rank detection now lives in `pamica/rank.py` and is called by the PyTorch, NumPy, and MLX backends. Only the *decision* is shared -- each backend still builds its sphering matrix in its own array library, so the bit-exact PyTorch path is not routed through a different eigensolver.

`pamica/tests/test_rank_policy.py` asserts all three size their models identically; PyTorch and NumPy run unconditionally, MLX via `importorskip`. That test is the anti-drift guard.

### Default changed to a relative floor
`mineig_rel=1e-12` is now the default; `mineig_rel=None` restores Fortran's absolute floor exactly.

The absolute floor is unit-dependent, and it is not only MEG that suffers. Average-referenced EEG -- an everyday step -- sits at `lambda_min/lambda_max = 8.5e-17` on the bundled sample, within a factor of ~10 of the 1e-15 floor, so the reference detects that rank deficiency essentially by luck. The relative floor detects it every time, and is scale-invariant across 1e-13 to 1e6.

Well-conditioned data is unaffected: real EEG sits at 5e-4, eight orders above the floor. Re-verified **bit-identical** against main after the refactor.

### Documentation
- **ADR 0004** -- the decision, the measurements, and the alternatives (including `max(mineig, rel)`, which was implemented first and is wrong: for Tesla-scale data the absolute term dominates and reproduces the bug).
- **`.rules/backend_parity.md`** -- new standing rule: no one-off backends, shared decisions go in shared modules, cross-backend tests mandatory, deliberate divergences must be recorded. Referenced from AGENTS.md Core Principles.
- **`docs/guides/amica-differences.md`** -- every deliberate pamica-vs-AMICA difference in one table, with why and how to restore reference behavior, plus a backend feature matrix. Added to the mkdocs nav.

The differences table is the answer to "which ones are different and why, in one look", and it is now the documented home for any future divergence.

Issue #240, opened 2026-08-16
numpy_impl share_comps produces NaN mu/beta and reports success


neuromechanist:
`numpy_impl` + `share_comps` produces NaN mixture parameters and reports success.

## Reproduction

```python
m = AMICA_NumPy(num_models=2, num_mix=3, max_iter=5, seed=7,
                share_comps=True, share_start=1, share_int=2)
m.fit(data)   # real sample EEG, 32ch x 4096
```

```
A:    nan=0/2048
mu:   nan=96/192      <-- half
beta: nan=96/192      <-- half
comp_used sum: 64 of 64
```

Two `RuntimeWarning: invalid value encountered in divide` fire during the fit, at `numpy_impl/core.py:1225` (`dmu_n/dmu_d`) and `:1231` (`dbeta_n/dbeta_d`). The fit then returns normally: `A` is clean, `gm` is clean, and nothing surfaces the NaNs.

## Why

Merging a column across models leaves the merged-away column with no responsibility mass, so its mixture denominators go to zero and the update is 0/0.

The PyTorch backend already handles exactly this: merged-away columns are frozen, "avoiding 0/0 NaN that Fortran tolerates behind its `comp_used` mask" (AGENTS.md, issue #60). Fortran survives it because `comp_used` masks the dead columns. In `numpy_impl` neither guard is present, and `comp_used` stays all-True, so nothing masks them and nothing reports them.

## Why it matters

This is a silent failure of the kind the project rules single out: a degenerate result presented as a successful fit. The `AMICA` wrapper's degenerate-fit contract (#50) is torch-only, so `AMICA_NumPy` has no equivalent refusal.

Not a regression. Found while adding `share_comps` coverage for #219; the gm-ordering change there touches `dAk` only and cannot affect `mu`/`beta`.

## Work

- [ ] Freeze merged-away columns in `numpy_impl` as the PyTorch backend does, or mask them via `comp_used` as Fortran does. Pick one and match a reference rather than inventing a third behavior.
- [ ] Whichever is chosen, a fit that ends with non-finite mixture parameters must not report success.
- [ ] Cross-backend test that `share_comps` leaves finite parameters in both backends (`.rules/backend_parity.md`).

Part of #231.

PR #241, opened 2026-08-16
Weight nd by pre-update model weights

### LINKED ISSUE #219: numpy_impl ndtmpsum uses current-iteration gm; Fortran uses the previous iteration's
Found in review of #217. Diagnostic-only, does not affect fitted parameters.

## The divergence

Fortran's `gm` is a module-level array that is not reassigned until `update_params` (`amica15.f90:1789`+), which runs **after** `accum_updates_and_likelihood` builds `dAk` (`:1731-1743`). So Fortran's `ndtmpsum` is computed with the **previous** iteration's `gm`.

In `numpy_impl`, `self.gm` is reassigned at `core.py:1124`/`1126`, before the `nd` block at `:1271-1288`. So ours uses the **current** iteration's `gm`.

## When it matters

Not for `num_models=1`, where `gm == 1`.

Not for the default disjoint `comp_list` multi-model case either: each column belongs to exactly one model, so the `gm[h]` numerator cancels exactly against `zeta[idx] = gm[h]`.

It matters only under `share_comps` with a genuinely shared column across two models, where the weighted average of directions is taken with different weights than Fortran would use.

## Scope

`self.nd` is a diagnostic and a convergence-check input; it is not used to update `A`, `mu`, or `beta`. So this cannot change a fitted decomposition. It can change when `use_grad_norm` fires, though `min_nd` is separately unreachable on small data (#218).

There is currently no test covering `numpy_impl` + `share_comps` + `nd`; the sharing tests are torch-only (`tests/torch_tests/test_ng_sharing.py`).

## Work

- [ ] Decide whether to snapshot `gm` before the mixture update in `numpy_impl` so the ordering matches Fortran, or to accept the deviation and document it as a known diagnostic-only difference.
- [ ] Add a `numpy_impl` + `share_comps` test that exercises `nd`, whichever way the above goes.
- [ ] Check whether `AMICATorchNG` has the same ordering question, since it computes `ndtmpsum` in the same region.

neuromechanist:
Closes #219.

## The ordering

Fortran builds `dAk` inside `accum_updates_and_likelihood` and only reassigns `gm` in `update_params`:

```fortran
dAk(:,comp_list(i,h)) = dAk(:,comp_list(i,h)) + gm(h)*dA(:,i,h)          ! ~1753
ndtmpsum = sqrt(sum(nd(iter,:),mask=comp_used) / (nw*count(comp_used)))  ! ~1761
gm = dgm_numer / dble(all_blks)                                          ! ~1788
```

So the reference weights by the *previous* iteration's model weights. Both backends reassigned `gm` first; both now snapshot it.

## Correction to the issue

#219 describes this as diagnostic-only. That holds for `numpy_impl`, where `dAk` feeds `nd` alone. It does **not** hold for `AMICATorchNG`, where the same `dAk` is the A-update (`self.A = self.A - self.lrate * dAk`), so the ordering reaches fitted parameters there. Worth stating plainly, since the issue's framing implied this PR could not change results.

## Measured effect

Single-model and Newton fits are bit-identical (`gm == 1`). Multi-model shifts by 1.0e-06 relative in log-likelihood.

That shift is not a semantic change: for a disjoint `comp_list` each column belongs to exactly one model, so `gm[h]` cancels against `zeta[idx] = gm[h]` and the result is algebraically identical either way. What changes is the floating-point association, amplified by the chaotic EM trajectory to ~1e-6 — the same order as the documented `block_size` trajectory shift, and well inside the ~5e-4 parity tolerance.

The weighting differs *semantically* only under `share_comps`, where a column belongs to more than one model and the cancellation no longer applies.

## Tests

Two added to `test_numpy_grad_norm.py`: one pinning the pre-update weighting with a forced shared column, one asserting a shared column keeps the gradient norm finite (`zeta` accumulates a positive weight per contributing model, so the division cannot go 0/0).

## Found on the way

Writing the sharing test surfaced #240: `numpy_impl` + `share_comps` leaves half of `mu` and `beta` NaN while reporting a successful fit. Pre-existing and independent of this change — the mixture update never reads `dAk` — so it is filed rather than fixed here.

### DIFF
--- a/pamica/numpy_impl/core.py
+++ b/pamica/numpy_impl/core.py
@@ -1350,6 +1350,11 @@ class AMICA:
         # num_models=1 (gm == 1) and for the default disjoint comp_list, where the
         # gm[h] factor cancels against zeta[idx]; it bites only under share_comps
         # with a genuinely shared column, where the two weightings differ.
+        #
+        # Diagnostic only *here*: this dAk feeds nd_value alone. The A-update below
+        # is a separate per-model loop that never reads dAk, so the ordering cannot
+        # move a fitted parameter in this backend. AMICATorchNG applies the same
+        # dAk to A, so there the identical fix does change fitted parameters.
         dAk = np.zeros_like(self.A)
         zeta = np.zeros(self.num_comps)
         for h in range(self.num_models):
--- a/pamica/numpy_impl/core.py
+++ b/pamica/numpy_impl/core.py
@@ -1183,6 +1183,13 @@ class AMICA:
             and self.comp_list is not None
             and self.A is not None
         )
+        # Fortran builds dAk from the model weights of the *previous* iteration:
+        # gm is not reassigned until update_params (amica15.f90:1789+), which runs
+        # after accum_updates_and_likelihood (:1731-1743). Snapshot it here so the
+        # nd block below weights by the same gm Fortran would (issue #219).
+        assert self.gm is not None
+        gm_prev = self.gm.copy()
+
         # Update model weights, normalizing by the number of samples the E-step
         # actually summed over: the good set under do_reject, else all samples.
         if self.do_reject:
@@ -1338,20 +1345,17 @@ class AMICA:
         # applied update lrate*dAk, and does not divide by lrate either
         # (amica15.f90:1742-1743) -- there is no missing factor here.
         #
-        # Fortran ordering caveat: Fortran's gm is not reassigned until
-        # update_params (amica15.f90:1789+), which runs AFTER dAk is built, so its
-        # ndtmpsum uses the previous iteration's gm. self.gm is already updated by
-        # this point. That is invisible for num_models=1 (gm == 1) and for the
-        # default disjoint comp_list, where the gm[h] factor cancels exactly
-        # against zeta[idx], but the two differ under share_comps with a genuinely
-        # shared column. Diagnostic only, it does not affect the fitted
-        # parameters; tracked in issue #219.
+        # Weighted by gm_prev, the pre-update model weights, because Fortran builds
+        # dAk before update_params reassigns gm (issue #219). Invisible for
+        # num_models=1 (gm == 1) and for the default disjoint comp_list, where the
+        # gm[h] factor cancels against zeta[idx]; it bites only under share_comps
+        # with a genuinely shared column, where the two weightings differ.
         dAk = np.zeros_like(self.A)
         zeta = np.zeros(self.num_comps)
         for h in range(self.num_models):
             idx = self.comp_list[:, h]
-            dAk[:, idx] += self.gm[h] * np.dot(directions[h].T, self.A[:, idx])
-            zeta[idx] += self.gm[h]
+            dAk[:, idx] += gm_prev[h] * np.dot(directions[h].T, self.A[:, idx])
+            zeta[idx] += gm_prev[h]
         nonzero = zeta > 0
         dAk[:, nonzero] /= zeta[nonzero]
         # comp_used is None until fit() sets it up, and the M-step is exercised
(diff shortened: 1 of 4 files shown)


neuromechanist:
## Review findings addressed

The test reviewer did the decisive thing I had not: it reverted the change and re-ran. **All tests still passed, on both backends.** The PR had no regression guard.

Three findings, all fixed:

**1. The ordering test was vacuous.** It mutated `comp_list` *after* `fit()` had returned, then asserted on `model.nd[-1]`, which was computed during that fit. The mutation was dead code with respect to every assertion, and the docstring described "two M-steps" that did not exist.

**2. Nothing would fail if the change were reverted.** The remaining assertions (`isfinite`, `> 0`, `gm` sums to 1) hold under either weighting.

**3. The PyTorch backend had no test at all**, despite being the backend where `dAk` is the A-update rather than a diagnostic — a direct gap against `.rules/backend_parity.md`.

## The discriminator

Hold one accumulator dict fixed across two M-steps, so the *post*-update gm (`dgm / n_samples`) is identical by construction, and vary only the gm present when `dAk` is built. Weighting by the post-update value makes the two runs identical; weighting by the pre-update value does not.

```python
nd_a = step_with_pre_update_gm(model, acc, n, [0.9, 0.1])
nd_b = step_with_pre_update_gm(model, acc, n, [0.1, 0.9])
assert nd_a != nd_b
```

Each has a complement asserting the same pre-update gm is deterministic, so they cannot pass on unrelated sensitivity, plus a single-model case pinning why issue #24 parity is untouched (`gm == 1`).

## Verified by reverting

- NumPy reverted -> `test_nd_uses_pre_update_model_weights` fails
- PyTorch reverted -> `test_ndtmpsum_uses_pre_update_model_weights` **and** `test_a_update_uses_pre_update_model_weights` fail

10 tests pass with the fix in place. New file `pamica/tests/torch_tests/test_ng_gm_ordering.py` covers the torch half, including that the fitted `A` itself moves with the ordering — the consequence unique to that backend.

neuromechanist:
## Second reviewer's findings addressed

The general code reviewer independently reproduced the revert-and-rerun result, and added three more.

**Comment accuracy — my regression.** The pre-PR numpy comment said *"Diagnostic only, it does not affect the fitted parameters"*; I dropped that clause, leaving a comment that reads identically to the torch one for a change with materially different consequences. Restored, and now states explicitly that numpy's A-update is a separate loop that never reads `dAk`, while torch applies the same `dAk` to `A`.

**Torch comment undersold its own scope.** It said "the ndtmpsum block below", but that block is also the A-update. Reworded to say `dAk` both drives the A-update and reports `ndtmpsum`. The reviewer noted this was inconsistent with the file's own `min_nd` docstring, which already describes `dAk` correctly as the update direction.

**Latent fragility.** `gm_prev = self.gm` aliased rather than copied, unlike numpy's `.copy()`. Safe today — `gm` is only ever rebound, never written in place — but an in-place write elsewhere would silently corrupt the snapshot. Now `.clone()`, with the reason recorded.

The reviewer also asked that the numpy/torch asymmetry be called out explicitly; it already is, under "Correction to the issue" in the PR description.

## Filed rather than fixed here

**#242** — the reviewer found that `numpy_impl`'s A-update is a per-model loop with no `gm` weighting, no `zeta` division and no post-merge A-freeze, so a genuinely shared column is updated **twice in sequence**, the second time against an already-updated `A`. Fortran applies one weighted average; the PyTorch backend matches Fortran. That is a different operation, not a rounding difference, and it means the two backends do not implement the same algorithm under `share_comps`.

Pre-existing and untouched by this PR (the loop is not in the diff), and out of scope here. With #240 it makes `numpy_impl` + `share_comps` a coherent piece of work rather than two loose ends.

Issue #242, opened 2026-08-16
numpy_impl share_comps applies a shared column's A-update twice


neuromechanist:
Found during the #241 review. Distinct from #240 (NaN `mu`/`beta`), same area: `numpy_impl`'s `share_comps` path is structurally further from Fortran than the `nd` ordering issue #219 addressed.

## Fortran

A single `dAk` is built as a `gm`-weighted average across contributing models, divided by `zeta = sum_h gm[h]`, then applied once (amica15.f90:1749-1761 build, `DAXPY` at :1807/:1814). A column shared by two models receives **one** averaged update.

## numpy_impl

The A-update is a separate per-model loop (`numpy_impl/core.py:1377-1381`):

```python
for h in range(self.num_models):
    idx = self.comp_list[:, h]
    self.A[:, idx] = self.A[:, idx] - self.lrate * np.dot(directions[h].T, self.A[:, idx])
```

No `gm` weighting, no `zeta` division, and no A-freeze guard — `grep` finds no `_a_frozen` equivalent in the file. `identify_shared_components` (`numpy_impl/utils.py:103-166`) genuinely can place one column index in `comp_list[:, h]` for two different `h`, so that column is updated **twice in sequence**, the second time against an already-updated `self.A`.

That is a different operation from Fortran's single weighted average, not a rounding difference.

## Contrast with the PyTorch backend

`AMICATorchNG` builds the `gm`-weighted `dAk`, divides by `zeta`, applies it once, and freezes A for ~6 iterations after a merge (issue #60). The two backends therefore do not implement the same algorithm under `share_comps`.

## Scope

- Default (disjoint `comp_list`) is unaffected: each column belongs to one model, so the loop applies exactly one update and there is nothing to average.
- Only reachable with `share_comps=True` *and* an actual merge.

## Work

- [ ] Port the `gm`-weighted single-application A-update to `numpy_impl`, matching Fortran and the PyTorch backend
- [ ] Port the post-merge A-freeze, or state why it is unnecessary here
- [ ] Cross-backend test that a shared column receives the same update in both (`.rules/backend_parity.md`)

Related: #240 (NaN mixture parameters on the same path), #219 (the `nd` ordering, fixed in #241). Part of #231.

PR #243, opened 2026-08-16
Fix comp_used staleness and NaN under sharing

### LINKED ISSUE #240: numpy_impl share_comps produces NaN mu/beta and reports success

neuromechanist:
Closes #240.

## Root cause

`identify_shared_components` rebuilt `comp_used = np.ones(...)` on every call. The merge loop skips any pair where `k1 == k2`, so once `comp_list` is already merged, no merge fires and the mask comes back **all-True while half the columns are dead**.

The unmasked mixture update then divided 0/0 for every merged-away column:

```
mu:   nan=96/192
beta: nan=96/192
comp_used: 64 of 64      <-- claims every column is live
```

and the fit returned normally.

## Fix

`comp_used` is now derived from the final `comp_list` — exactly the set of referenced columns — matching `AMICATorchNG.comp_used`, which is a property derived the same way and so cannot go stale.

The mixture update now skips unused columns and freezes them at their last finite value, as the torch backend already did. Fortran carries NaN there harmlessly behind its own `comp_used` mask; keeping them finite means a fit cannot report success while holding NaN parameters.

```
mu:   nan=0/192
beta: nan=0/192
comp_used: 32 of 64
```

`np.errstate` guards the discarded branch of each `np.where`, so the 0/0 that is computed and thrown away no longer warns every iteration.

## Parity

Default (disjoint `comp_list`) is **bit-identical** for single- and multi-model — verified before/after on real sample EEG. Every change is behind the `used` mask, which is all-True without a merge.

## Tests

`pamica/tests/test_numpy_share_comps.py`, 5 tests. Verified by reverting `comp_used` to `np.ones(...)`: `test_comp_used_matches_the_columns_comp_list_references` and `test_sharing_leaves_finite_mixture_parameters` both fail.

## What is deliberately not here

#242 — the shared-column A-update, where numpy takes one step per contributing model instead of Fortran's single weighted application — is **not** fixed in this PR.

I implemented it (a one-line change to use the `dAk` already computed above, identical to the torch backend), then wrote three tests for it: a magnitude comparison against private columns, a cross-backend comparison against `AMICATorchNG`, and a finiteness check. **All three passed equally with the per-model loop and with the single weighted application** — none distinguished the two. The magnitude test's premise was simply wrong: a merged column legitimately moves ~3x a private one, because the averaged direction need not be small.

Rather than ship a numerical change to the A-update with tests that give false confidence — the exact failure caught in #241's review — the change is reverted and #242 stays open until it has a test that fails on revert. The comment at the loop records why.

### DIFF
--- a/pamica/numpy_impl/core.py
+++ b/pamica/numpy_impl/core.py
@@ -1222,14 +1222,35 @@ class AMICA:
                     self.iter,
                 )
 
-        # Update mixture weights
-        self.alpha = updates["dalpha_n"] / np.sum(updates["dalpha_n"], axis=0)
+        # A component merged away by share_comps is no longer referenced by
+        # comp_list, so no sufficient statistic accumulates into its column and
+        # the divisions below are 0/0 = NaN. Update only used columns and freeze
+        # the rest at their last finite value, as AMICATorchNG does (Fortran
+        # carries the NaN harmlessly behind its comp_used mask; keeping them
+        # finite means a fit cannot report success while holding NaN parameters,
+        # issue #240). All-True with the default comp_list, so the ordinary path
+        # is unchanged.
+        used = (
+            self.comp_used
+            if self.comp_used is not None
+            else np.ones(self.num_comps, dtype=bool)
+        )
+
+        # Update mixture weights. errstate because np.where evaluates both
+        # branches: an unused column's 0/0 is computed and discarded, and would
+        # otherwise warn on every iteration after a merge.
+        with np.errstate(invalid="ignore", divide="ignore"):
+            alpha_next = updates["dalpha_n"] / np.sum(updates["dalpha_n"], axis=0)
+        self.alpha = np.where(used, alpha_next, self.alpha)
 
         # Exact-EM mixture location/scale (Fortran :1978/:1993). These are
         # fixed-point updates -- mu += dmu_n/dmu_d, beta *= sqrt(dbeta_n/dbeta_d)
         # -- NOT first-order gradient steps, so they carry no lrate.
-        self.mu = self.mu + updates["dmu_n"] / updates["dmu_d"]
-        self.beta = self.beta * np.sqrt(updates["dbeta_n"] / updates["dbeta_d"])
+        with np.errstate(invalid="ignore", divide="ignore"):
+            mu_next = self.mu + updates["dmu_n"] / updates["dmu_d"]
+            beta_next = self.beta * np.sqrt(updates["dbeta_n"] / updates["dbeta_d"])
+        self.mu = np.where(used, mu_next, self.mu)
+        self.beta = np.where(used, beta_next, self.beta)
         self.beta = np.clip(self.beta, self.invsigmin, self.invsigmax)
         # Fortran keeps a live "NaN in sbeta!" canary here (amica17.f90:1996-2000);
         # the exact-EM mu/beta divisions are unguarded (matching Fortran), so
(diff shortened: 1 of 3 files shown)


neuromechanist:
## Review findings addressed

The reviewer ran the revert matrix per change. Unlike #241, this PR's tests **do** go red on revert — but it found a subtler version of the same blind spot, plus a real gap I introduced.

**1. Two tests skipped rather than failed under the bug.** Their guards (`if used_first.all(): skip`, `if not unused.any(): skip`) were keyed to precisely the symptom the bug produces — a stale all-True mask reads as "no merge fired". So under the bug they silently opted out instead of failing. Same blind spot as #241, wearing a skip instead of a pass.

Fixed by forcing the merge deterministically (`_force_collinear_pair` copies one model's column onto another's, so a merge must fire) and turning both guards into assertions. Revert matrix now: **4 of 5 fail**, where before it was 2 failing and 2 skipping.

**2. `alpha` lost its only signal.** There is no `np.isfinite(self.alpha)` check anywhere in the file — only `mu`/`beta` have one. My `np.errstate` silences the discarded 0/0 for dead columns, but it equally silenced numpy's warning for a *genuine* 0/0 in a **live** column, which was the sole indication it had happened. The NaN would still surface eventually through the nan-LL restart, but with the origin lost — cutting against this same function's stated philosophy for the `rho` reset two blocks down.

Added an explicit `alpha` finiteness check with a warning, mirroring the `mu`/`beta` canary.

**3. The freeze comment overclaimed.** It said unused columns are frozen at their last finite value, but `rho` is excluded — its own `1e-8` floor keeps it finite, so it drifts rather than freezing. Harmless, since no dead column is read downstream, but the comment now says so.

## Confirmed, not changed

- `comp_used` derivation is correct on first call, on repeat calls, and for never-merged columns — `comp_list` starts as a bijection and merges only overwrite, so `np.unique` always recovers the live set.
- Freezing vs Fortran's NaN-tolerant mask is inert: `get_unmixing_matrices` and the A-update both index through `comp_list`, which never references a merged-away column.
- The fix silently corrects a second consumer for free: `nd_value` reads `comp_used` and was over-counting its denominator with dead columns after a merge, deflating the reported gradient norm.

Two low-confidence notes I am leaving as-is: a `num_comps > data_dim * n_models` edge case inherited verbatim from `AMICATorchNG.comp_used`, and the absence of a cross-backend `comp_used` agreement test. Neither is introduced here; the second is worth adding when #242 lands, since that PR needs a cross-backend test anyway.

PR #245, opened 2026-08-16
Document mir_history offset; fix waypoint assertion

neuromechanist:
Closes #161, and repairs a test my own #241 broke.

## #161 was mostly already done

Both of its testable claims already have tests in `test_ng_convergence.py` — `test_mir_history_survives_keep_best_restore` and `test_mir_history_empty_after_save_load`. Only the third, informational item was outstanding.

Verified before documenting it: `ll` comes from the E-step accumulator at the top of the loop, `_update_parameters` runs, and the MIR waypoint is computed after. So `mir_history_[i]` reflects the parameters *after* iteration `i`'s update while `ll_history[i]` is the likelihood of the parameters *before* it. Zipping them by index compares different parameter states. Now stated on both the backend attribute and the `AMICA.mir_history_` docstring.

## A regression from #241

Running the whole file rather than a subset surfaced `test_mir_history_survives_keep_best_restore` failing on `dev`:

```
assert 50 == (55 - 1)
```

I bisected it: the test passes at `257d20e` (the commit before #241 merged) and fails at `c80fb33` (the merge). **#241 caused it.** That PR's gm-ordering change shifts multi-model trajectories by ~1e-6, which moved the `min_dll` stop to iteration 55; MIR waypoints land on multiples of `mir_step=5`, so the last one is at 50, not 54.

The assertion `last_it == len(ll_history) - 1` was fragile from the start — it only held while the stop happened to coincide with a waypoint, and any trajectory perturbation could break it. It now asserts what the test is actually about: the last waypoint is the highest multiple of `mir_step` at or below the final iteration, and is within one interval of the end, so a restore that truncated the trajectory would still fail it.

## Why CI did not catch it

The test is not `slow`-marked, so CI runs it, and CI was green on #241's branch. I have not established why the branch and the squashed merge behave differently here; the practical lesson is that I ran a subset locally on #241 and would not have seen this either way. Flagging rather than leaving it implicit.

### DIFF
--- a/pamica/amica.py
+++ b/pamica/amica.py
@@ -88,6 +88,9 @@ class AMICA:
         the mid-fit ``W``/``sphere``. Like ``ll_history_``, a ``keep_best``
         restore does not rewrite it -- use :meth:`mir` on the fitted model for
         the value of the *returned* parameters, not ``mir_history_[-1]``.
+        Not index-aligned with ``ll_history_``: entry ``i`` is computed after
+        iteration ``i``'s update, while ``ll_history_[i]`` is the likelihood of
+        the parameters before it, so the two are one update apart (issue #161).
 
     Examples
     --------
--- a/pamica/tests/torch_tests/test_ng_convergence.py
+++ b/pamica/tests/torch_tests/test_ng_convergence.py
@@ -765,7 +765,15 @@ def test_mir_history_survives_keep_best_restore(real_data):
 
     Uses the same genuine-overshoot recipe as
     ``test_keep_best_restores_genuine_overshoot_under_min_dll_stop`` above,
-    with ``mir_step=5`` added."""
+    with ``mir_step=1`` added.
+
+    ``mir_step=1`` is load-bearing, not incidental. The ``min_dll``/``maxincs``
+    stop halts one or two iterations past the peak, so with any coarser step
+    the last waypoint lands *before* the best iterate and the window a
+    truncating restore would damage is never sampled -- a restore that dropped
+    every waypoint after the best iterate would leave this fixture unchanged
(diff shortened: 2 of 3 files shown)


neuromechanist:
Review findings addressed; merging.

**Finding 1 (Critical) — the assertion could not witness a truncating restore.** Confirmed and fixed. The reviewer's sabotage (filter `mir_history_` to `t[0] <= best_it` right after `_restore_params`) passed under the old assertion because `min_dll`/`maxincs` halts 1-2 iterations past the peak while `mir_step=5`, so the last waypoint (50) sat before both `best_it` (53) and `final_it` (54) — the damaged window was never sampled. Fixed by making `mir_step=1` load-bearing: every iteration is a waypoint, so one always lands strictly inside `(best_it, final_it]`. Re-ran the same sabotage against the new test: **it now fails**, and passes again after revert.

**Finding 2 (Important) — `assert last_it > final_it - 5` was a tautology.** Agreed, the algebra is exact. Removed rather than reworded, and replaced with a count assertion (`len(mir_history_) == final_it + 1`) that catches partial truncation the last-entry check would miss.

**Finding 3 (Important) — magic `5` duplicated three times.** Dissolved by the fix: with `mir_step=1` the assertions are `last_it == final_it` and `len(...) == final_it + 1`, so no literal remains.

**Finding 4** — noted; the new form is robust to stopping-iteration drift, since it no longer depends on where the stop lands relative to a waypoint boundary.

The docstring accuracy confirmation was useful independently: the off-by-one being documented is real (`ll_history[i]` pre-update, `mir_history_[i]` post-update).

22/22 in the file pass, `ty` clean.

PR #247, opened 2026-08-16
Add a macOS Apple Silicon CI job

### LINKED ISSUE #246: CI runs Linux only: MLX backend and the Apple Silicon numerical path are never tested
## Problem

Every job in `.github/workflows/ci.yml` is `runs-on: ubuntu-latest` — there is no
platform matrix. Two consequences:

1. **`pamica/tests/mlx_tests/` has never run in CI.** MLX is the recommended Apple
   GPU path per `AGENTS.md` (~15-25 ms/it, ~7x torch-CPU, faster than an RTX 4090
   at EEG scale) and #81 extended it to multi-model. It is covered only by local
   runs on a maintainer's machine.
2. **The Accelerate-vs-OpenBLAS numerical path is untested**, so macOS-only
   failures are structurally invisible.

Consequence 2 is not hypothetical. `test_mir_history_survives_keep_best_restore`
was green on #241's PR CI and failed immediately after the merge locally. The
cause was not a stale base (#241's base was one version-bump commit old): the
~1e-6 trajectory shift moved the stopping iteration from 55 to 54, and the
assertion only held when the stop landed on a multiple of `mir_step`. On Linux
the stop landed elsewhere and CI never saw it. Adding CI on `dev` pushes would
*not* have caught this — only a macOS runner would.

Four more files gate on darwin/native-engine and are silently skipped in CI:
`test_native_engine.py`, `test_fortran_adapter.py`, `test_rank_policy.py`,
`test_dimsweep_threads.py`.

## Proposed fix

Add a `macos-26` job (macOS 26, arm64, GA). Standard runners are **free and
unmetered for public repositories**, and pAMICA is public, so this costs nothing.
Note `macos-14` is deprecated and its plain label is Intel x64; `macos-latest`
currently points to `macos-26`. Avoid `-large`/`-xlarge` variants, which are
billed even for public repos.

Scope to decide:
- Full suite on macOS, or a targeted job (MLX + native engine + the
  numerically-sensitive convergence tests)?
- Whether to also add `push: branches: [dev]` so post-merge drift on the
  integration branch is caught (worth doing, but note it would not have caught
  the failure above).

## Explicitly not proposed

A self-hosted runner on a maintainer's Mac. `sccn/pAMICA` is public, and GitHub
advises self-hosted runners only for private repos: a fork PR can execute
arbitrary code on the runner, which is not sandboxed between jobs. The hosted
`macos-26` image covers the same ground with no exposure. Self-hosted remains
the only option for CUDA, which no hosted runner provides.

neuromechanist:
Closes #246 (first half: the MLX/Accelerate gap).

## Why

Every job in `ci.yml` was `ubuntu-latest`. Two things were therefore untested:

- **`pamica/tests/mlx_tests/` had never run in CI.** MLX is the recommended
  Apple GPU backend per `AGENTS.md`; `pyproject.toml` even excludes
  `pamica/mlx_impl/*` from coverage with the comment that CI "cannot exercise
  it and always measures 0%".
- **The Accelerate BLAS numerical path.** Not theoretical: the
  `test_mir_history_survives_keep_best_restore` assertion was green on #241's
  PR CI and failed immediately after merge on macOS. The cause was not a stale
  base — the ~1e-6 trajectory shift moved the stopping iteration from 55 to 54,
  and the assertion only held when the stop landed on a multiple of `mir_step`.
  Linux never saw it. Adding CI on `dev` pushes would not have caught this
  either; only a macOS runner would.

## What

One `test-macos` job on `macos-26` (current arm64 image; `macos-14` is arm64 too
but deprecated). Standard runners are free and unmetered for public repos, so
this costs nothing — deliberately not a `-large`/`-xlarge` variant, which are
billed even here.

Verified locally on macOS 26.6.1 arm64 before pushing:
- `mlx.core.__version__` -> `0.32.0`, so the import assertion is real
- `LAPACK_LIBS="-framework Accelerate" bash native/build.sh` builds cleanly
- `pamica/tests/mlx_tests/` collects 11 tests and runs

The MLX import is asserted explicitly rather than left to `mlx_tests`'
self-skip: a silent skip would leave this job green while testing nothing,
which is the exact failure mode it exists to prevent.

`--no-cov` because the Linux job owns the coverage gate and the Codecov upload;
a second partial run would report a different total for the same commit.

## Deliberately not in scope

- **Enabling the `slow` Fortran-parity tests on macOS.** They drive the bundled
  x86_64 `amica15mac`, which needs Rosetta on an arm64 runner. Pointing them at
  the arm64 binary this job already builds is the natural follow-up; left in
  #246.
- **`push: branches: [dev]`.** Worth doing, but a separate concern, and note it
  would not have caught the failure described above.
- **A self-hosted runner.** `sccn/pAMICA` is public and GitHub advises
  self-hosted runners only for private repos (a fork PR can run arbitrary code
  on an unsandboxed runner). Hosted `macos-26` covers the same ground with no
  exposure. Self-hosted remains the only option for CUDA.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,6 +113,50 @@ jobs:
           files: coverage.xml
           fail_ci_if_error: false
 
+  # Apple Silicon coverage (issue #246). The Linux jobs above cannot exercise
+  # two things the project depends on:
+  #   * pamica/tests/mlx_tests/ -- MLX is Apple-GPU only and is the recommended
+  #     Apple backend (AGENTS.md), so it had never run in CI at all.
+  #   * the Accelerate BLAS numerical path. Summation order differs from
+  #     OpenBLAS enough to move a stopping iteration, which is how a broken
+  #     assertion in test_ng_convergence.py stayed green on Linux through #241
+  #     and failed immediately on macOS.
+  # macos-26 is the current arm64 image (macos-14/15 are also arm64; 14 is
+  # deprecated). Standard runners are free and unmetered for public repos --
+  # do not switch to a -large/-xlarge variant, which is billed even here.
+  test-macos:
+    name: Test (macOS 26, Apple Silicon)
+    needs: [lint, typecheck]
+    runs-on: macos-26
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.12"
+          enable-cache: true
+      # --extra mlx is the whole point of this job: without it mlx_tests/
+      # self-skips and the run adds nothing the Linux job did not already do.
+      - run: uv sync --extra mlx
+      - name: Assert MLX is importable
+        # A silent skip here would make this job green while testing nothing,
+        # which is the failure mode it exists to prevent.
+        run: uv run python -c "import mlx.core; print('mlx', mlx.core.__version__)"
+      - name: Build native AMICA binary (for engine E2E tests)
+        # Same recipe as the macos-arm64 target in release-binaries.yml:
+        # gfortran from brew's gcc, LAPACK/BLAS from the Accelerate framework.
+        run: |
+          brew install gcc
+          LAPACK_LIBS="-framework Accelerate" bash native/build.sh
+          echo "PAMICA_NATIVE_BINARY=$PWD/native/amica15_shim" >> "$GITHUB_ENV"
+      # "not slow" matches the Linux job: the slow tests drive the bundled
+      # x86_64 amica15mac reference binary, which needs Rosetta on an arm64
+      # runner. Pointing them at the arm64 binary built above is the follow-up
+      # tracked in #246, not part of closing the MLX/Accelerate gap.
+      # --no-cov: the Linux job owns the coverage gate and the Codecov upload;
+      # a second partial run would report a different total for the same commit.
+      - name: pytest (excluding slow / Fortran-binary parity tests)
+        run: uv run pytest -m "not slow" -n auto --no-cov
+
   test-mne:
     name: Test MNE wrapper (optional [mne] extra)
     needs: [lint, typecheck]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,11 @@ jobs:
     name: Test (macOS 26, Apple Silicon)
     needs: [lint, typecheck]
     runs-on: macos-26
+    env:
+      # The only job on hardware where torch auto-selects MPS (torch_impl/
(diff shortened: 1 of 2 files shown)


neuromechanist:
Findings addressed in da41ebd. One correction on the Critical, with evidence.

**Critical 1 — "the MLX tests will very likely still self-skip": empirically not the case.** I checked the completed run's log rather than reasoning about it. On the hosted `macos-26` runner, all 11 MLX tests **PASSED**:

```
PASSED pamica/tests/mlx_tests/test_mlx_backend.py::test_sufficient_stats_match_numpy_reference
PASSED pamica/tests/mlx_tests/test_mlx_backend.py::test_converged_ll_matches_torch_float32
PASSED pamica/tests/mlx_tests/test_mlx_backend.py::test_multimodel_matches_torch_float32
... (11 total)
============ 402 passed, 10 skipped, 1 xfailed in 582.91s (0:09:42) ============
```

The 10 skips are `test_fortran_adapter.py` (3) and `test_ng_pdf_families.py` (5) plus two others — no MLX test among them. So this runner does expose Metal to MLX and `mx.default_device().type == gpu` holds. The cited upstream reports are real but do not describe this image.

**That said, the finding is right as insurance and I have implemented it.** You correctly spotted the second gate at `test_mlx_backend.py:28-33` that my import assertion did not cover — if a future runner image loses GPU passthrough, the job would go green with 11 silent skips. The step now asserts `mx.default_device().type == mx.DeviceType.gpu`.

**Important 2 — the `-m "not slow"` rationale was wrong.** Confirmed both halves. `test_sample_data.py`'s slow tests never invoke a binary, and the two in `test_ng_backend.py` (:1096, :1179) pass `binary_path=None`, which resolves through `PAMICA_NATIVE_BINARY` **first** (`native/resolver.py:112`) — already set by this job to the arm64 binary it builds. Rosetta was never involved. Comment corrected to say they are excluded for wall-clock, and that enabling them would give CI its first Fortran parity coverage.

**Important 3 — `PYTORCH_ENABLE_MPS_FALLBACK`.** Added. This is indeed the only job whose hardware can auto-select MPS.

**Important 4 — `EXTRA_LDFLAGS`.** Added `-static-libquadmath` so the recipe genuinely matches `release-binaries.yml`. Verified the build locally with both flags: arm64 Mach-O, clean.

**Suggestion — stale coverage comment.** Updated to state why the omission is still *required* (this job runs `--no-cov`, so the total still comes only from the ubuntu job) rather than leaving a claim a future reader would find false.

**Suggestion — `release-binaries.yml` still on `macos-14`.** Good catch on the November 2 2026 EOL; filing separately rather than widening this PR.

Issue #248, opened 2026-08-16
MLX backend has no convergence stops: every fit runs to max_iter


neuromechanist:
## What

`pamica/mlx_impl/core.py` implements **neither** convergence stop. Verified by
grep — zero references to any of:

`min_dll`, `use_min_dll`, `maxincs`, `min_nd`, `use_grad_norm`, `ndtmpsum`

Its only `stop_reason` values are `max_iter` (:570), `lrate_floor` (:659) and the
degenerate ones (`nan_ll`/`singular_ll` :590, `nan_params` :634). So an MLX fit
runs the full iteration budget unless the learning rate floors or the fit
degenerates.

The other backends stop on likelihood stagnation (`min_dll`, on by default at
`1e-9`) long before `max_iter` on real data — measured at iteration 326-1076 on
the bundled sample depending on the BLAS build. The same nominal configuration
therefore does materially different work on MLX than on PyTorch/NumPy/Fortran.

## Why it matters

Unsupported parameters fail loudly today — the constructor rejects unknown
kwargs with `TypeError`, and `do_newton=True`/non-GG `pdftype` raise
`NotImplementedError`. But *this* gap is silent: nothing signals that the fit
had no convergence criterion. A user moving a working config from the PyTorch
backend to MLX for the documented Apple-GPU speedup (~15-25 ms/it, ~7x
torch-CPU, `AGENTS.md`) gets a fit that quietly ignores stopping and burns the
whole budget.

It also makes MLX unable to participate in convergence-behavior claims. #218
documents that `min_nd` is unreachable on small recordings "in any of the three
implementations"; MLX cannot be evaluated on that axis at all, because it never
computes the gradient norm.

## Scope

1. **`min_dll` / `use_min_dll` / `maxincs`** — the consecutive-small-gain rule
   (amica15.f90:1060-1072). The likelihood is already computed per iteration, so
   this is the cheap half and the one that changes real runs.
2. **`min_nd` / `use_grad_norm`** — requires accumulating `ndtmpsum` (Fortran
   amica15.f90:1731-1743). Note the gradient norm is also a useful diagnostic in
   its own right (epic #231), independent of the stop.
3. Decide whether `lrate_floor`'s missing `.or. (ndtmpsum .le. min_nd)` half
   (the `grad_norm_floor` stop) comes along with (2).

Per `.rules/backend_parity.md` this should land with cross-backend tests
asserting MLX stops at the same iteration as the PyTorch backend on the same
[...]

PR #249, opened 2026-08-16
Record the MLX convergence-stop divergence

neuromechanist:
Documentation half of #248 (the implementation stays open).

## What

`docs/guides/amica-differences.md`'s backend table covered Newton, PDF families,
component sharing, outlier rejection, precision and rank detection — but not the
convergence stops. MLX implements **neither**:

| grep target | hits in `pamica/mlx_impl/core.py` |
|---|---|
| `min_dll`, `use_min_dll`, `maxincs` | 0 |
| `min_nd`, `use_grad_norm`, `ndtmpsum` | 0 |

Its only `stop_reason` values are `max_iter` (:570), `lrate_floor` (:659) and the
degenerate ones (:590, :634). **An MLX fit always runs the full iteration
budget**, while the other backends stop on `min_dll` (on by default at `1e-9`) at
iteration 326-1076 on the bundled sample.

## Why this one needs saying

The doc previously closed with "MLX limitations raise `NotImplementedError`
rather than differing silently." Verified by probing the constructor — that is
true for what it covers:

| passed to `AMICAMLXNG(...)` | result |
|---|---|
| `do_newton=True` | `NotImplementedError` |
| `min_dll`, `min_nd`, `use_grad_norm`, `keep_best`, `do_reject` | `TypeError` (no such parameter) |

But a stop that does not exist has no parameter to reject, so nothing signals
its absence. That is the gap the section now names, with the practical
consequence: pick `max_iter` for MLX from where the other backends actually
converge on your data, not from their default budget.

It also scopes a claim in `validation.md` — "`min_nd` is unreachable in all
three implementations" means Fortran, PyTorch and NumPy. MLX never computes the
quantity, so it cannot be evaluated on that axis.

## Also corrected

The `keep_best` / MIR / persistence rows were added and verified per backend
rather than assumed — NumPy has no `keep_best`, no `mir`, and no `state_dict`
(it persists through the EEGLAB `amicaout` format instead), so an earlier draft
calling it "partial" would have been wrong.

Docs-only; no code paths touched.

### DIFF
--- a/docs/guides/amica-differences.md
+++ b/docs/guides/amica-differences.md
@@ -121,7 +121,7 @@ Separate from reference divergences: the optional MLX backend is a subset.
 | Precision | f64/f32 | f64 | f32 only | f64 |
 | Rank detection | yes | yes | yes | yes (absolute floor) |
 | `min_dll` stop | yes | yes | **no** | yes |
-| `min_nd` stop / gradient norm | yes | yes | **no** | yes |
+| `min_nd` stop / gradient norm | yes | yes (as `min_grad_norm`) | **no** | yes |
 | `keep_best` best-iterate restore | yes | no | no | n/a |
 | MIR diagnostic | yes | no | no | n/a |
 | Persistence | `state_dict` | EEGLAB `amicaout` | none | EEGLAB `amicaout` |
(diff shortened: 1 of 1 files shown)


neuromechanist:
Both Critical findings confirmed and fixed in f2ca8d7. They had a single root cause, correctly identified: this section was drafted against #244's `validation.md` rewrite, which is not on `dev`.

**Critical 1 — cross-reference to content that does not exist on the target branch.** Correct. The phrase "in all three implementations" lives only on `218-document-min-nd`. Rather than sequence the two PRs, I made this section self-contained: it no longer asserts what another page says, so merge order no longer matters. Verified there is now no reference to `validation.md`, "three implementations", or the 326-1076 figure in this file.

**Critical 2 — the 326-1076 framing implied MLX is uniquely budget-bound.** Correct, and worth stating plainly: this is the same error #244 exists to fix, reintroduced one file over. PyTorch reaches its own `max_iter=100` default long before `min_dll` can fire, so "MLX always runs the full budget" was not a contrast at all. Rewritten around the distinction that actually holds — what *raising* `max_iter` buys you: on the other backends a larger budget lets the likelihood stop end the fit once it stops improving; on MLX a larger budget is simply spent. No measurement is cited now, so nothing here depends on an unmerged source.

**Important 4 — `min_nd` vs `min_grad_norm`.** Added in both the table row and the prose, so a `numpy_impl` user grepping for `min_nd` is not left concluding the feature is absent rather than renamed.

Thanks for checking the per-backend cells independently rather than taking the table's word — the `keep_best`/MIR/persistence rows were ones I'd verified after an earlier draft got NumPy wrong, and it is useful to have that confirmed from a second read.

Issue #251, opened 2026-08-17
Support MNE `bad_*` annotations in `AMICAICA.fit()`


sunyuhongwr:
I found one remaining MNE-compatibility issue related to bad annotations. Currently, `AMICAICA.fit()` obtains the fitting data through:
X = picked.get_data(**range_kw)
but does not appear to account for `raw.annotations` containing segments whose descriptions start with `bad`.
In MNE-Python, `ICA.fit()` uses `reject_by_annotation=True` by default, so annotated bad segments are excluded from the data used for ICA fitting.
For example:
```python
raw.set_annotations(
    mne.Annotations(
        onset=[100.0],
        duration=[10.0],
        description=["bad_artifact"],
    )
)

ica = mne.preprocessing.ICA(...)
ica.fit(raw)
```
The 10-second bad segment is not used for the ICA decomposition. With the current `AMICAICA.fit()`, however, the corresponding samples appear to still be passed to AMICA. This could potentially affect the decomposition because strongly contaminated segments can contribute substantially to the AMICA likelihood optimization.

### Why this matters for MNE compatibility

This also creates an inconsistency with the MNE-facing API.
For example, after fitting:
```python
model_probability = ica.get_model_probability(raw)
```
the returned model probabilities still contain values for the annotated bad segments.
I understand that this is not necessarily wrong for `get_model_probability()` itself—the method is evaluating the fitted model on the supplied data. The problem is that, from an MNE user's perspective, the model was expected to be fitted while excluding the `bad_*` segments, just as `mne.preprocessing.ICA.fit()` does.

### Possible implementation
Would it be possible to add something analogous to MNE's:
```python
reject_by_annotation=True
```
to `AMICAICA.fit()`?
For example:
```python
amicaica.fit(
    raw,
    reject_by_annotation=True,
)
```
Internally, the wrapper could construct a mask of valid samples from the `bad_*` annotations and fit AMICA only on the valid samples.
Importantly, I don't think the bad segments necessarily need to be physically deleted from the `Raw` object. A cleaner approach may be:
1. identify samples covered by `bad_*` annotations;
2. create a boolean mask of good samples;
3. extract only the good samples for AMICA fitting;
4. concatenate the remaining good segments along time;
5. preserve the original `Raw` object and annotations.
This would allow AMICA to be trained only on valid data while preserving the original time axis and annotations for subsequent MNE operations.

### One additional consideration
[...]

PR #252, opened 2026-08-19
Add reject_by_annotation support to AMICAICA.fit and AMICAICA._data_for

psysunyuhong:
## Background

This PR addresses [Issue #251](https://github.com/sccn/pAMICA/issues/251).

When using `AMICAICA` with MNE `Raw` data, segments marked with `bad_*` annotations should not be used during AMICA fitting, consistent with the behavior of MNE's `ICA.fit()`.
Previously, `AMICAICA.fit()` always used the complete Raw data, including samples covered by `bad` annotations. This could cause artifact-contaminated segments to be included during AMICA model fitting.

## Changes

This PR adds `reject_by_annotation` support to `AMICAICA.fit()`.
- Added `reject_by_annotation=True` as a fitting parameter.
- When enabled, samples covered by annotations whose description starts with `"bad"` are excluded from AMICA fitting.
- The original Raw sample mask is stored in `good_sample_mask_`.
- The mask preserves the correspondence between the fitted data and the original Raw sample timeline.
- When `reject_by_annotation=False`, all samples are retained, preserving the previous behavior.

## Model Probability Reconstruction

The stored sample mask also allows per-sample AMICA model probabilities to be mapped back to the original Raw timeline.
For example, if a 30-second recording contains 600 rejected samples:
- Original Raw data: 3000 samples
- Samples used for AMICA fitting: 2400 samples
- Rejected samples: 600 samples

The model probabilities can therefore be reconstructed to the original 3000-sample timeline, with rejected samples represented as `NaN`.
This preserves the temporal correspondence required for subsequent time-resolved model weighting and reconstruction.

## Testing

A test was added using an artificial 30-second MNE `Raw` object with three `bad_*` annotation segments.

The test verifies that:

- `reject_by_annotation=True` excludes the expected samples.
- `reject_by_annotation=False` retains all samples.
- `good_sample_mask_` has the correct length and number of rejected samples.
- Model probabilities can be mapped back to the original Raw timeline.
- Rejected samples are represented as `NaN`.
- Good-sample probabilities are preserved after reconstruction.

All tests pass. Test file available in pamica/tests/mne_tests/test_bad_annotation.py
Test on 306-channel Elekta data with around ~70 rank, test pass.
Related to #251

### DIFF
--- a/pamica/mne_compat/core.py
+++ b/pamica/mne_compat/core.py
@@ -29,6 +29,7 @@ import torch
 # base type-check/CI env has no mne, hence the scoped ignore (issue #139).
 import mne  # ty: ignore[unresolved-import]
 from mne.preprocessing import ICA as _MNEICA  # ty: ignore[unresolved-import]
+from mne.annotations import _annotations_starts_stops
 
 from ..amica import AMICA
 from ..torch_impl import PDFTYPE_NAMES
@@ -164,6 +165,8 @@ class AMICAICA:
         self.ch_names_: Optional[list] = None
         self.n_components_: Optional[int] = None
         self.pre_whitener_: Optional[np.ndarray] = None
+        self.reject_by_annotation_: bool = False
+        self.good_sample_mask_: Optional[np.ndarray] = None
         self.converged_: bool = False
         self.stop_reason_: Optional[str] = None
         self._n_samples: Optional[int] = None
@@ -180,6 +183,7 @@ class AMICAICA:
         picks=None,
         start: Optional[int] = None,
         stop: Optional[int] = None,
+        reject_by_annotation: bool = True,
         **fit_kwargs,
     ) -> "AMICAICA":
         """Fit AMICA to the data of an MNE ``Raw`` or ``Epochs``.
@@ -196,6 +200,10 @@ class AMICAICA:
         start, stop : int | None
             Sample range for ``Raw`` input, passed to ``get_data``; ``None`` uses
             the full recording. Not supported for ``Epochs`` (raises).
+        reject_by_annotation : bool, default=True
+            If ``True``, omit time segments whose annotation description starts
+            with ``"bad"`` when fitting from :class:`~mne.io.Raw`. This matches
+            MNE's ``ICA.fit`` behavior. Has no effect for :class:`~mne.BaseEpochs`.    
         **fit_kwargs
             Forwarded to :meth:`AMICA.fit` (e.g. ``max_iter``, ``lrate``,
             ``do_newton``) and the backend constructor (e.g. ``block_size``).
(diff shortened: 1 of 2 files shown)


neuromechanist:
@claude, can you do pr review here?

neuromechanist:
Thanks @psysunyuhong -- this is a well-aimed contribution and the `good_sample_mask_` idea is exactly right. I reviewed it, kept your design, and pushed a rework commit (95dd55f) directly to your branch, plus a merge of `dev` (now the repository default branch and this PR's base). What the review found and how the rework addresses it:

## Bugs fixed

1. **`start`/`stop` composition**: the mask stayed `True` outside the fitted `[start, stop)` range, so `mask.sum()` disagreed with the number of fitted samples and the reconstruction workflow broke whenever a sample range was used. The mask is now `False` outside the range (test: `test_start_stop_composes_with_rejection`).

2. **Epochs path**: `Epochs.get_data(reject_by_annotation=None)` raises `TypeError` -- MNE's `Epochs.get_data` has no such kwarg (only `Raw`'s does), so every Epochs fit would have crashed. Restored the plain call (test: `test_epochs_fit_is_unaffected`).

3. **Private import**: `mne.annotations._annotations_starts_stops` is private API and also failed the `ty` CI job. The mask is now derived from the public `get_data(reject_by_annotation="omit", return_times=True)` -- the same call that produces the fitted array, so mask and data cannot disagree by construction (test: `test_mask_indexes_exactly_the_fitted_columns`).

## Semantics adjusted

Your `_data_for` applied the stored fit-time mask to whatever instance is passed, which shortens `get_model_probability` output (while `get_sources` stays full-length) and forces consumers to re-index manually -- your test did exactly that dance. MNE's own precedent (`ICA.score_sources(..., reject_by_annotation=True)`) rejects by the *passed* instance's annotations at evaluation time. So now `get_model_probability`, `plot_model_probability`, `mir`, and `pmi` all take `reject_by_annotation=True`, reject by the passed instance's own `bad_*` annotations, and per-sample outputs keep the full `n_times` timeline with `NaN` in rejected columns. That is the "Model Probability Reconstruction" section of your PR description built into the library: `ica.get_model_probability(raw)` returns the full-timeline array with NaN over the bad segments directly, no manual mapping, and it stays correct when scoring a different `Raw` than was fitted.

`good_sample_mask_` and `reject_by_annotation_` are kept as fitted attributes recording exactly what the fit consumed, as you proposed.

## Tests

Rewritten as a pytest suite on the bundled real sample EEG (project policy: no synthetic-data tests; the script form also executed two AMICA fits at import time and printed FAIL without failing the build). The new suite additionally pins that fitting an annotated Raw equals fitting AMICA directly on the kept columns (`test_fit_equals_manual_column_drop`) and that NaN samples inside a bad-annotated span are tolerated (dropped before the finite check), which matters for your MEG use case.

All 64 MNE-wrapper tests pass locally, ruff/ty clean. Once CI is green I will squash-merge. Thanks again for the report and the implementation -- and for offering to test on your Elekta data.

